PR #12, opened 2024-12-03
Add some initial timing/rhythm features

HuwCheston:
This PR adds two new modules to `musmart`: `musmart.timing` and `musmart.rhythm`. Currently the following features are implemented:

- `beat_upbeat_ratio`
- `npvi`
- `tempo_slope`
- `tempo_drift`
- `tempo_fluctuation`

All features have examples in the docstring which should work with `doctest`, although I'm not sure how you wanted this implemented @pmcharrison. Happy to migrate these to separate unit tests inside `tests` if we want to do that instead.


musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+    ValueError
+        If too few consecutive beats are provided
+
+    Examples
+    --------
+    >>> my_upbeats = np.array([0.6, 1.4])
+    >>> my_beats = np.array([0., 1., 2.])
+    >>> returned = beat_upbeat_ratio(my_beats, my_upbeats)
+    >>> [round(r, 2) for r in returned]
+    [0.67, 1.5]
```
pmcharrison, line 100: I'm confused about the ordering. I would have expected 1.5 to come first and 0.67 to come second? As I understand it, the first downbeat has length 0.6 and the first upbeat has length 0.4, and 0.6 / 0.4 is 1.5.

pmcharrison: Maybe the problem is that it depends which upbeat you compare with which downbeat? This needs to be made clearer in the docstring.

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+        https://doi.org/10.1525/mp.2021.38.4.372
+
+    """
+
+    # Parse beats and upbeats to an array
+    beats = np.array(beats)
+    upbeats = np.array(upbeats)
+    # Remove any values from our `upbeats` variable that are also contained in `beats`
+    upbeats_ = upbeats[~np.isin(upbeats, beats)]
+    beats_ = beats[~np.isin(beats, upbeats)]
```
pmcharrison, line 129: `isin` is rarely the right operation if you are dealing with floating point numbers. You should define a tolerance for equality

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+    # Assert that there is no overlap between both arrays
+    assert not any(np.isin(beats_, upbeats_)), "`beats` and `upbeats` must be exclusive but they overlap"
```
pmcharrison, line 131: You did not state that overlap wasn't allowed in the docstring.

musmart/timing/tempo_modeling.py
```diff
@@ -0,0 +1,146 @@
...
+from typing import Iterable
+
+import numpy as np
+
+
+__all__ = ["tempo_slope", "tempo_drift", "tempo_fluctuation"]
+__author__ = "Huw Cheston"
+
+
+def _fit_tempo_linregress(beats: Iterable):
```
pmcharrison, line 15: As above, put the top-level function at the top of the script.

pmcharrison, line 15: Specify what type the Iterable should contain

musmart/timing/tempo_modeling.py
```diff
@@ -0,0 +1,146 @@
...
+    """Fits linear regression of BPM measurements vs onset time"""
+
+    from scipy.stats import linregress
+    # Dependent variable: BPM measurements
+    y = np.array(60 / np.diff(beats))
```
pmcharrison, line 20: I feel like converting beat timestamps into instantaneous tempo should be a separate function, then this function should be solely occupied with computing tempo slope from instantaneous tempo

pmcharrison, line 20: `np.diff(beats)` will give weird results if any beats are skipped, it would probably be wise to mention that missing values are not permitted in the docstring

musmart/timing/tempo_modeling.py
```diff
@@ -0,0 +1,146 @@
...
+    # Predictor variable: the onset time
+    x = np.array(beats[1:])
+    # Handling NaNs: remove missing values from both arrays
```
pmcharrison, line 23: I don't think you should silently remove NaNs. Why would they occur? Presumably when `np.diff(beats)` returns 0. But why would that happen?

musmart/timing/tempo_modeling.py
```diff
@@ -0,0 +1,146 @@
...
+    Examples
+    --------
+    >>> stable = [1., 2., 3., 4.]
+    >>> unstable = [1., 2.1, 2.9, 3.5]
+    >>> tempo_drift(stable) < tempo_drift(unstable)
+    True
+
+    """
+
+    return float(_fit_tempo_linregress(beats).stderr)
```
pmcharrison, line 110: I don't see how the standard error of the regression line can be described as "the gradient of the slope of signed overall tempo change in a performance"....? These seem like quite different concepts to me.

musmart/timing/tempo_modeling.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def tempo_fluctuation(beats: Iterable) -> float:
```
pmcharrison, line 113: add type to Iterable

musmart/timing/tempo_modeling.py
```diff
@@ -0,0 +1,146 @@
...
+    ----------
+    [1] Cheston, H., Schlichting, J.L., Cross, I., and Harrison, P.M.C. (2024).
+        Jazz Trio Database: Automated Annotation of Jazz Piano Trio Recordings Processed Using
+        Audio Source Separation. Transactions of the International Society for Music Information
+        Retrieval, 7/1 (pp. 144–158). https://doi.org/10.5334/tismir.186
+
+    """
+
+    inter_beat_intervals = np.diff(beats)
+    return float(np.nanstd(inter_beat_intervals) / np.nanmean(inter_beat_intervals))
```
pmcharrison, line 146: How come here you are computing SD of IBIs but in the docstring you say you are computing SD of tempos? They're not the same thing.

pmcharrison, line 146: Why are you silently dropping NaNs? Why would you have NaNs here?

musmart/rhythm/npvi.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
```
pmcharrison, line 1: Please can you use Jupytext format rather than ipynb for your notebook?

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+    use_log2_burs : bool, optional
+        Whether to use the log_2 of inter-onset intervals to calculate BURs, as employed in [2]. Defaults to False.
+    allow_multiple_burs_per_beat : bool, optional
+        If True, calculates a beat-upbeat ratio for every upbeat between two consecutive beats. Otherwise,
+        calculates a BUR only for the first upbeat. Defaults to False.
+
+    Returns
+    -------
+    list
+        The calculated BUR values.
```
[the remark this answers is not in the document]
HuwCheston: To be discussed: whether we want to use `typing.List` or just the inbuilt `list` here. Will open an issue...

musmart/timing/swing.py
pmcharrison: I see how this came from a suggestion from Mark..... but to be honest, I don't see why it's useful to have a special case for this situation. Maybe because it's ambiguous in that case which beat it corresponds to? But that depends on your matching algorithm, which I'm not clear on yet.

MarkGotham: Thanks both.

My concern is to convey clearly to users what should be passed to the parameters and the relationship between them. 

Is is all event positions in one, and those assigned the role of `beats` in the other? If so, the expectation of using a strict subset suggests that it is sensible to include an `assert` check.

Depending on the algorithm (which I'm also not yet totally clear on – but intend to look through closely soon) and range of intended use cases (which is probably still open to debate), we could leave it to the user to decide whether they need/want a check on this? E.g., `check_all_in: bool = False`

Wrt tolerance for equality, perhaps try one of the `isclose` options, @HuwCheston ?

pmcharrison: Yes, I think that we need clarification/discussion of the input parameters and the performed algorithm, and then this should shed light on the present question. In general though I find that silent skipping behaviour is dangerous as it often hides important problems from the user. Better to raise a warning or error, normally.

pmcharrison: Presumably though if you had an 'upbeat' happening exactly on the beat that would imply that there was something wrong with your data collection

examples/analyze_jazz_timing.ipynb
```diff
@@ -0,0 +1,555 @@
+{
```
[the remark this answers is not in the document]
MarkGotham: ... Or split the notebook examples to a different PR so we can merge the main functionality

musmart/rhythm/npvi.py
```diff
@@ -0,0 +1,47 @@
...
+
+    Returns
+    -------
+    float
+        The extracted nPVI value.
+
+    Examples
+    --------
+    >>> res = normalized_pairwise_variability_index(np.array([1, 2, 3]))
+    >>> round(res)
```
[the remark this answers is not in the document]
MarkGotham: +1

musmart/rhythm/npvi.py
```diff
@@ -0,0 +1,47 @@
...
+__author__ = "Huw Cheston"
+
+
+def normalized_pairwise_variability_index(sequence: Iterable) -> float:
+    """
+    Extracts the normalised pairwise variability index (nPVI).
+
+    The nPVI is a measure of variability between successive elements in a sequence, commonly used
+    in music analysis to quantify rhythmic variability [1].
+
```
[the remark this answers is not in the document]
MarkGotham: Here's a first go at the nPVI (normalised pairwise variability index) equation in LaTeX and applied to this sphinx environment:
```
The nPVI (normalised pairwise variability index) equation is
.. math:: nPVI = \frac{100}{(n-1)} \sum_{i=1}^{n-1} \left| \frac{(d_i - d_{i+1})}{((d_i + d_{i+1})/2)} \right|
   :label: nPVI
where :math:`n` is the number of intervals, and :math:`d_i` is the duration of the :math:`i^{th}` interval. 
```
Hope this helps a bit. Please do check. E.g., I haven't tried it in situ yet. Just getting the ball rolling.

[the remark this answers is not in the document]
MarkGotham: ^^ tagging @HuwCheston @pmcharrison

musmart/rhythm/npvi.py
MarkGotham: ... and again consider moving the notebook to a separate PR

musmart/rhythm/npvi.py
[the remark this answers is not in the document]
HuwCheston: Referring to the original paper (https://www.jstor.org/stable/10.1525/mp.2013.31.1.10, appendix), I think the equation should look something like

```
 \text{nPVI} = \frac{100}{m-1} \times \sum\limits_{k=1}^{m-1} \left| \frac{d_k - d_{k+1}}{\frac{d_k + d_{k+1}}{2}} \right|
 ```

Tested in an online LaTeX editor and looks good:

image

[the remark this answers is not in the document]
Comment (HuwCheston) [reply musmart/rhythm/npvi.py:21]: Resolved in https://github.com/music-computing/amads/pull/83

File: musmart/rhythm/npvi.py (hunk 13 above)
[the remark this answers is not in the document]
Comment (HuwCheston) [reply musmart/rhythm/npvi.py:35]: Resolved in https://github.com/music-computing/amads/pull/83

File: musmart/rhythm/npvi.py [hunk 15]
```diff
@@ -0,0 +1,47 @@
...
+    Returns
+    -------
+    float
+        The extracted nPVI value.
+
+    Examples
+    --------
+    >>> res = normalized_pairwise_variability_index(np.array([1, 2, 3]))
+    >>> round(res)
+    53
```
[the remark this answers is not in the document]
HuwCheston: Resolved in https://github.com/music-computing/amads/pull/83 --- added two examples from the literature that provide "answers" we can compare our implementation to

musmart/rhythm/npvi.py
```diff
@@ -0,0 +1,47 @@
...
+
+    References
+    ----------
+    [1] Daniele, J. R., & Patel, A. D. (2013). An Empirical Study of Historical Patterns in Musical Rhythm.
+        Music Perception, 31/1 (pp. 10-18). https://doi.org/10.1525/mp.2013.31.1.10
+
+    """
+
+    numerator = sum([abs((k - k1) / ((k + k1) / 2)) for (k, k1) in zip(sequence, sequence[1:])]) * 100
+    denominator = sum(range(len(sequence))) - 1
```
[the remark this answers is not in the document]
HuwCheston: Resolved in https://github.com/music-computing/amads/pull/83

musmart/rhythm/npvi.py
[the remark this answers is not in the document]
MarkGotham: Out of interest what was the difference?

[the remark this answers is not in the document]
MarkGotham: Just m vs n and k vs i?

musmart/timing/tempo_modeling.py
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/85

HuwCheston: Fixed in https://github.com/music-computing/amads/pull/85

musmart/timing/tempo_modeling.py
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/85

HuwCheston: Fixed in https://github.com/music-computing/amads/pull/85

musmart/timing/tempo_modeling.py
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/85

musmart/timing/tempo_modeling.py
HuwCheston: Yep, my fault, we're just interested in the standard error of the slope. Fixed in https://github.com/music-computing/amads/pull/85

musmart/timing/tempo_modeling.py
HuwCheston: In https://github.com/music-computing/amads/pull/85

musmart/timing/tempo_modeling.py
HuwCheston: Yep, meant tempos (as implied by `tempo_fluctuation`), fixed in https://github.com/music-computing/amads/pull/85

HuwCheston: Fixed in https://github.com/music-computing/amads/pull/85, we explicitly raise an error on NaN values

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+        ioi2 = beat2 - upbeat_bw
+        # Calculate the beat-upbeat ratio
+        bur = ioi2 / ioi1
+        # Express with base-2 log if required
+        if use_log2_burs:
+            bur = log2(bur)
+        yield float(bur)
+
+
+def beat_upbeat_ratio(
```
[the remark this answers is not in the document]
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/87

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+        beats: Iterable,
```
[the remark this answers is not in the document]
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/87

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+    ("long") eighth note beat by the second ("short") beat. A BUR value of 2 indicates "perfect" swing
+    (e.g., a triplet quarter note followed by a triplet eighth note), while a BUR of 1 indicates "even"
+    eighth-note durations.
+
+    Accepts two iterables of timestamps, corresponding to `beats` and `upbeats`. These lists should be sorted,
+    but missing values can be included and will be handled internally in the function.
+
+    By default, the function assumes there is exactly one upbeat between every two consecutive beats. If there
+    are multiple upbeats between two beats, only the first upbeat is used for BUR calculation unless
+    `allow_multiple_burs_per_beat=True`, in which case a BUR is calculated for every upbeat.
```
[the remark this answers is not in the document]
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/87. I agree that a `multiple upbeats` situation isn't realistically swing. However, I do think it is beneficial for a user to pass in inputs where multiple upbeats may occur (or, indeed, cases where a beat has no upbeats). So I've added in logic to skip over these cases while maintaining the structure of the output.

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+        # Express with base-2 log if required
+        if use_log2_burs:
+            bur = log2(bur)
+        yield float(bur)
+
+
+def beat_upbeat_ratio(
+        beats: Iterable,
+        upbeats: Iterable,
+        use_log2_burs: bool = False,
```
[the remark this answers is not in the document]
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/87

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+
+    Returns
+    -------
+    list
+        The calculated BUR values.
+
+    Raises
+    ------
+    ValueError
+        If too few consecutive beats are provided
```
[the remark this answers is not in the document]
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/87

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+        calculates a BUR only for the first upbeat. Defaults to False.
+
+    Returns
+    -------
+    list
+        The calculated BUR values.
+
+    Raises
+    ------
+    ValueError
```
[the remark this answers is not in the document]
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/87

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+        If too few consecutive beats are provided
+
+    Examples
+    --------
+    >>> my_upbeats = np.array([0.6, 1.4])
```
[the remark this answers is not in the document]
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/87

musmart/timing/swing.py
```diff
@@ -0,0 +1,141 @@
...
+    >>> my_beats = np.array([0., 1., 2.])
+    >>> returned = beat_upbeat_ratio(my_beats, my_upbeats)
+    >>> [round(r, 2) for r in returned]
```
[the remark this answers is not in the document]
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/87 by removing doctests and just adding a separate pytest file

musmart/timing/swing.py
HuwCheston: This was just a mistake I think! Fixed in https://github.com/music-computing/amads/pull/87

musmart/timing/swing.py
HuwCheston: Should be fixed in https://github.com/music-computing/amads/pull/87. My thinking is along the lines with Peter's, that if we have an upbeat colliding exactly (or near exactly) with a beat, we should raise an error. There's a validation function `_validate_bur_inputs` that checks for these and other edge cases.

musmart/timing/swing.py
HuwCheston: Fixed in https://github.com/music-computing/amads/pull/87, now stated in docstring

PR #14, opened 2024-12-03
Added LZ77 complexity algorithm

HuwCheston:
This PR adds the complexity measure based on the LZ77 compression algorithm from [Cheston, Schlichting, Cross, & Harrison (2024)](https://royalsocietypublishing.org/doi/10.1098/rsos.240920). Input is a discrete 1D `numpy` array with any dtype, which is converted internally into a string representation based on the number of unique elements. Output is the length of the compressed string after applying LZ77.


musmart/algorithm/complexity.py
```diff
@@ -0,0 +1,127 @@
...
+    Here we calculate the complexity of a discrete `sequence` (could be pitch classes, binned inter-onset intervals)
+    by converting it into a string representation, compressing using LZ77 [1], then calculating the length of the
+    compressed string as in [2]. Higher values mean that more information is required to represent the sequence,
+    indicating greater complexity.
+
+    Currently, only sequences with fewer than 96 unique elements are supported.
+
+    Parameters
+    ----------
+    sequence : np.ndarray
```
pmcharrison, line 54: The types in the docstring should match the type hints in the source code. Or you could just delete the former.

MarkGotham:
Nudge @HuwCheston to keep this on the radar.

Are you up for finishing this off? 

Looks like a small and uncontroversial PR edit ... and you have first refusal on implementing as it's your PR ;)

musmart/algorithm/complexity.py
```diff
@@ -0,0 +1,87 @@
...
+    Returns:
+        int: the length of the compressed string
+
+    Examples:
+        NB. some examples are taken from https://timguite.github.io/jekyll/update/2020/03/15/lz77-in-python.html
+
+        >>> lz77(np.array(["a"]))
+        1
+
+        >>> lz77(np.array(["a", "b", "a"]))
```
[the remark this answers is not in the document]
HuwCheston: Resolved in https://github.com/music-computing/amads/pull/14/commits/3cdcd889f4c77a7900fc1b8d69414faae1a4429c

musmart/algorithm/complexity.py
HuwCheston: Resolved in https://github.com/music-computing/amads/pull/14/commits/3cdcd889f4c77a7900fc1b8d69414faae1a4429c

HuwCheston:
Ok, added a few new commits that should allow any sequence of immutable objects to be passed in. Also added a decode method. Tested and works with e.g. strings `"abababab"`, lists of strings, floats, etc. Think that covers the use cases we'll probably want.

pmcharrison (approved):
Great job!

PR #27, opened 2024-12-04
Added Step Contour from FANTASTIC toolbox

Dav8Circle:
Implemented the Step Contour features from FANTASTIC (Müllensiefen, 2009) https://www.doc.gold.ac.uk/isms/m4s/FANTASTIC_docs.pdf

Takes an input list of pitches (ints) and durations (floats) and normalizes them to an array containing 64 ints. These ints are the pitch values, repeated proportionally to their duration in the inputs. Script contains multiple functions capturing features of this contour representation.

TODO: fix long lines in a way that doesn't break the doctest


HuwCheston (dismissed):
Made a few general comments --- haven't looked fully at the mathematical detail, but seems generally sound

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,247 @@
...
+    (64 tatums total).
+
+    Parameters
+    ----------
+    durations : list of float
+        List of duration values measured in tatums
+
+    Returns
+    -------
+    list of float
```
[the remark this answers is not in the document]
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,247 @@
...
+    ValueError: The length of pitches must be equal to the length of durations.
+
+    References
+    ----------
+    Müllensiefen, D. (2009). FANTASTIC: Feature ANalysis Technology Accessing
+    STatistics (In a Corpus): Technical Report v1.5
+    https://www.doc.gold.ac.uk/isms/m4s/FANTASTIC_docs.pdf
+    """
+    if len(pitches) != len(durations):
+        raise ValueError("The length of pitches must be equal to the length of durations.")
```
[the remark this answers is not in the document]
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,247 @@
...
+
+    Throws a ValueError where input lists are different in length
+    >>> step_contour([60, 62, 63, 65, 72], [1, 1, 4, 4]) # doctest: +ELLIPSIS
+    Traceback (most recent call last):
+      ...
+    ValueError: The length of pitches must be equal to the length of durations.
+
+    References
+    ----------
+    Müllensiefen, D. (2009). FANTASTIC: Feature ANalysis Technology Accessing
```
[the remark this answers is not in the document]
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,247 @@
...
+        List of length 64 containing repeated pitch values
+    """
+    result = []
+    for pitch, duration in zip(pitches, normalized_durations):
+        # Round the duration to nearest integer and repeat the pitch that many times
+        repetitions = round(duration)
+        result.extend([pitch] * repetitions)
+
+    # Ensure the result is exactly 64 elements long
+    if len(result) > 64:
```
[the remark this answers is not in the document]
Dav8Circle: fixed

pmcharrison (commented):
Some initial comments

musmart/melody/contour/step_contour.py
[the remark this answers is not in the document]
HuwCheston: +1 for function parameter (I wasn't really sure what this meant in context anyway). My preference would be:

- Global constant defined at top of file
- Function signature that uses this global constant as default argument

e.g., 
```
CONSTANT = 64
def func(constant: int = CONSTANT):
    ...
```

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,247 @@
...
+    Flat contour
+    >>> flat = [60] * 64
+    >>> step_contour_global_direction(flat)
+    0.0
+    """
+    if len(contour) != 64:
+        raise ValueError("The length of the step contour must be 64.")
+
+    corr = np.corrcoef(contour, np.arange(64))[0, 1]
+    if np.isnan(corr):
```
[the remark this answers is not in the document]
Dav8Circle: You're right. This implementation was too simplistic, aimed at dealing with the case where the input list contains all the same values `[60, 60, 60, ..., 60]` - this by definition causes a division by zero problem, and `np.corrcoef` returns `NaN` as a result. I've fixed it now with a case to catch repeated values in input, and if an input is invalid, it will instead return `None`, rather than a value that could be construed as a correlation coefficient.

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,247 @@
...
+    ...           72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72,
+    ...           72, 72, 72, 72, 72, 72, 72, 72, 72, 72]
+    >>> step_contour_local_variation(contour)
+    0.19047619047619047
+    """
+    if len(contour) != 64:
+        raise ValueError("The length of the step contour must be 64.")
+
+    # Calculate the mean absolute difference between adjacent values
+    local_variation = sum(abs(contour[i + 1] - contour[i])
```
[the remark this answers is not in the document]
Dav8Circle: yeah this is lots clearer, added now, thanks

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,247 @@
...
+    Throws a ValueError when contour length is not 64
+    >>> step_contour_global_variation([60, 62, 63]) # doctest: +ELLIPSIS
+    Traceback (most recent call last):
+      ...
+    ValueError: The length of the step contour must be 64.
+    """
+    if len(contour) != 64:
+        raise ValueError("The length of the step contour must be 64.")
+
+    return float(np.std(contour))
```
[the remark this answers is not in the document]
Dav8Circle: better safe than sorry! fixed

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,247 @@
...
+Calculates the Step Contour of a melody, along with related features, as implemented
+in the FANTASTIC toolbox of Müllensiefen (2009).
+"""
+
+__author__ = "David Whyatt"
+
+import numpy as np
+
+
+def _normalize_durations(durations: list[float]) -> list[float]:
```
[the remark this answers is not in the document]
Dav8Circle: fixed!

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,161 @@
...
+            return 0.0
+
+        try:
+            corr = np.corrcoef(
+                self._contour,
+                np.arange(self._step_contour_length)
+            )[0, 1]
+            return float(corr)
+        except Exception:
+            return None
```
HuwCheston, line 214: CC'ing @pmcharrison. I'm not sure about this from an API perspective. I feel fairly strongly that if we have an error, this shouldn't be skipped silently.

I understand the returning 0.0 for a flat array and think this makes sense intuitively. But I would allow `np.corrcoef` to fail rather than returning None here.

HuwCheston: In a use case scenario, if I'm running this class as part of a pipeline where only one input will fail, it's possible to imagine how I could get weird errors from e.g. trying to perform further `numpy` operations on an array that contains both `floats` and `NoneType` values. This would waste time, when really the root of the problem is the operation failing.

pmcharrison: Totally agree with @HuwCheston, never write `except Exception` unless you then reraise it later. Your linter should catch it. @Dav8Circle if you don't have it already, can you install the Pylint extension on your Cursor?

musmart/melody/contour/step_contour.py
```diff
@@ -1,260 +1,161 @@
...
+        durations: list[float]
+    ) -> list[int]:
+        """Calculate the step contour from input pitches and durations."""
+        normalized_durations = self._normalize_durations(durations)
+        return self._expand_to_vector(pitches, normalized_durations)
+
+    @property
+    def contour(self) -> list[int]:
+        """Get the step contour vector."""
+        return self._contour
```
pmcharrison, line 169: Why do we need both `_contour` and `contour?

Dav8Circle: This was suggested by cursor - the motivation is that `_contour` is immutable so won't be overwritten by the user after initialisation (either in error or on purpose). Then `contour` gives a public interface for the encapsulated `_contour`

I don't have strong feelings about it, but it feels like good practice!

musmart/melody/contour/step_contour.py
Dav8Circle: My mistake, I get you now - definitely better to just let `np.corrcoef` fail than try and catch in a confusing way. @pmcharrison I've installed Pylint as requested, hadn't set it up in this IDE!

musmart/melody/contour/step_contour.py
pmcharrison: I see! No, I'd prefer to remove this, if we did that everywhere it would make things very verbose.

Dav8Circle: Understood! fixed

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+            List of length step_contour_length containing repeated pitch values
+
+        Examples
+        --------
+        >>> sc = StepContour([60], [1.0], step_contour_length=4)
+        >>> sc._expand_to_vector([60, 62], [2.0, 2.0])
+        [60, 60, 62, 62]
+        """
+        result = []
+        for pitch, duration in zip(pitches, normalized_durations):
```
pmcharrison, line 132: The implementation of `_expand_to_vector` is succinct but not quite satisfying to me. The calculation of repetitions via rounding can accumulate errors, and then you need this hacky adjustment at the end that distorts the end of the melody. 

Here's an implementation that, while admittedly a bit more complex / less readable, avoids these distortions I think. It would be worth checking though.

```py
        if abs(sum(normalized_durations) - self._step_contour_length) > 1e-6:
            raise ValueError(
                f"The sum of normalized_durations ({sum(normalized_durations)}) must "
                f"be equal to the step contour length ({self._step_contour_length})"
            )
        # We interpret the output list as a vector of pitch samples taken
        # at times 0, 1, 2, ..., 63 where 63 = _step_contour_length - 1
        # and the length of the normalized melody is 64.

        output_length = self._step_contour_length
        output_pitches = [None for _ in range(output_length)]
        output_times = list(range(output_length))

        output_index = 0
        offset = 0.0

        # Iterate over the input pitches and durations
        for sounding_pitch, duration in zip(pitches, normalized_durations):
            offset += duration

            while True:
                # Step through the output list, and populate any time slots that
                # are occupied by the current note.
                if output_index >= output_length:
                    break
                output_time = output_times[output_index]
                if output_time >= offset:
                    break
                output_pitches[output_index] = sounding_pitch
                output_index += 1

        return output_pitches
```

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Returns
+        -------
+        list[int]
+            List of length step_contour_length containing repeated pitch values
+
+        Examples
+        --------
+        >>> sc = StepContour([60], [1.0], step_contour_length=4)
+        >>> sc._expand_to_vector([60, 62], [2.0, 2.0])
```
pmcharrison, line 128: To make this more elegant you could consider making `_expand_to_vector` a class method and adding `step_contour_length` to its parameter list. Then you could write

```py
>>> StepContour._expand_to_vector([60, 62], [2.0, 2.0], step_contour_length=4)
[60, 60, 62, 62]
```

That I think would be preferable because you don't need to create a StepContour object to use the function.

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+        ]
+        return normalized
+
+    def _expand_to_vector(
+        self,
+        pitches: list[int],
+        normalized_durations: list[float]
+    ) -> list[int]:
+        """Helper function to create a vector of length step_contour_length by repeating
+        each pitch value proportionally to its normalized duration.
```
pmcharrison, line 111: I think if you make the changes I suggest below then you would update the docstring to something like "Helper function that resamples the melody to length step_contour_length."

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+        >>> sc = StepContour([60, 62], [1.0, 1.0], step_contour_length=4)
+        >>> sc._calculate_contour([60, 62], [1.0, 1.0])
+        [60, 60, 62, 62]
+        """
+        normalized_durations = self._normalize_durations(durations)
+        return self._expand_to_vector(pitches, normalized_durations)
+
+    @property
+    def global_variation(self) -> float:
+        """Calculate the global variation of the step contour.
```
pmcharrison, line 161: You should provide a complete mathematical definition in the docstring

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Returns
+        -------
+        float
+            Float value representing the global variation of the step contour
+
+        Examples
+        --------
+        >>> sc = StepContour([60, 62, 64], [1.0, 1.0, 1.0])
+        >>> round(sc.global_variation, 2)
```
pmcharrison, line 171: I have made an update to the doctest config (pull the latest changes to this branch) so that you should be able remove the `round` from this

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+        1.64
+        """
+        return float(np.nanstd(self.contour))
```
pmcharrison, line 174: I would not silently skip NANs -- I think if there is a NAN then there is some kind of error that needs detecting, no?

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+0.1111...
+"""
+
+__author__ = "David Whyatt"
+
+import numpy as np
+
+
+class StepContour:
+    """Class for calculating and analyzing step contours of melodies.
```
pmcharrison, line 25: Does this analysis assume that the melody does not contain rests? We should say so.

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+        --------
+        >>> sc = StepContour([60, 62, 64], [1.0, 1.0, 1.0])
+        >>> round(sc.global_variation, 2)
+        1.64
+        """
+        return float(np.nanstd(self.contour))
+
+    @property
+    def global_direction(self) -> float:
+        """Calculate the global direction of the step contour.
```
pmcharrison, line 178: You should provide a complete mathematical definition in the docstring

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+        Returns
+        -------
+        float
+            Float value representing the global direction of the step contour
+            Returns 0.0 if the contour is flat
+
+        Examples
+        --------
+        >>> sc = StepContour([60, 62, 64], [1.0, 1.0, 1.0])
+        >>> round(sc.global_direction, 3)
```
pmcharrison, line 189: Can remove round now

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+        0.943
+        """
+        if len(set(self.contour)) == 1:
```
pmcharrison, line 192: It's a bit dangerous to use `set` here for comparing floats. I think it would be better to remove this logic, calculate corr as you have below, then check if the answer is nan. If the answer is nan, and length > 1, then the global direction is indeed 0.

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
+"""Calculates the Step Contour of a melody, along with related features, as implemented
+in the FANTASTIC toolbox of Müllensiefen (2009) [1].
+
```
pmcharrison, line 3: Does Müllensiefen provide any reference for the original source of this algorithm?

musmart/melody/contour/step_contour.py
pmcharrison, line 189: Round can be removed now

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+        -------
+        float
+            Float value representing the global direction of the step contour
+            Returns 0.0 if the contour is flat
+
+        Examples
+        --------
+        >>> sc = StepContour([60, 62, 64], [1.0, 1.0, 1.0])
+        >>> round(sc.global_direction, 3)
+        0.943
```
pmcharrison, line 190: Where did you get 0.943 from? What makes you think this is right answer? Looking at the input, one might have thought the output should be 1.0....

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+
+    @property
+    def global_direction(self) -> float:
+        """Calculate the global direction of the step contour.
+
+        Returns
+        -------
+        float
+            Float value representing the global direction of the step contour
+            Returns 0.0 if the contour is flat
```
pmcharrison, line 184: You should add a test case for this

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Examples
+        --------
+        >>> sc = StepContour([60, 62, 64], [1.0, 1.0, 1.0])
+        >>> round(sc.global_direction, 3)
+        0.943
+        """
```
pmcharrison, line 191: You should also add a test case for a descending melody

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+            return 0.0
+        corr = np.corrcoef(
+            self.contour,
+            np.arange(self._step_contour_length)
+        )[0, 1]
+        return float(corr)
+
+    @property
+    def local_variation(self) -> float:
+        """Calculate the local variation of the step contour.
```
pmcharrison, line 202: Need to add mathematical definition of this

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+        Returns
+        -------
+        float
+            Float value representing the local variation of the step contour
+
+        Examples
+        --------
+        >>> sc = StepContour([60, 62, 64], [1.0, 1.0, 1.0])
+        >>> sc.local_variation
+        0.06349206349206349
```
pmcharrison, line 213: Too many decimal places

pmcharrison:
Some good stuff, some things to follow up on though!

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
+"""Calculates the Step Contour of a melody, along with related features, as implemented
```
pmcharrison, line 1: The package has now been reorganised so that musmart is called amads. When you are ready to do so, you should merge master into your branch and then move the file accordingly. I would probably wait until you have addressed the other comments though as moving the file might mess up the comment alignment.

musmart/melody/contour/step_contour.py
Dav8Circle: I've validated this with the original implementation in R - because step contour is exactly that, stepped, the correlation with a linearly increasing function is made slightly weaker!

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Examples
+        --------
+        >>> sc = StepContour([60], [1.0])
+        >>> sc._normalize_durations([2.0, 2.0])
+        [32.0, 32.0]
+        """
+        total_duration = sum(durations)
+        if total_duration == 0:
+            return durations
```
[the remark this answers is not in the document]
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
Dav8Circle: Added

musmart/melody/contour/step_contour.py
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
Dav8Circle: Fair enough, had this discussion with Huw a while ago, I suppose there's nothing gained by letting things progress if there are NaNs in the calculation - changed back to `return float(np.std(self.contour))`

musmart/melody/contour/step_contour.py
Dav8Circle: Rests are treated as extending the duration of the previous note - will add to docstring now

musmart/melody/contour/step_contour.py
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
Dav8Circle: done!

musmart/melody/contour/step_contour.py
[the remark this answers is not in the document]
Dav8Circle: Added and validated against original R implementation

musmart/melody/contour/step_contour.py
Dav8Circle: Okay, have reimplemented this - we do throw some warnings now with pytest when we allow divisions by zero/None, but as you say, we catch it so it's probably fine

musmart/melody/contour/step_contour.py
Dav8Circle: cited now

musmart/melody/contour/step_contour.py
Dav8Circle: done

musmart/melody/contour/step_contour.py
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
Dav8Circle: done

musmart/melody/contour/step_contour.py
Dav8Circle: fixed

musmart/melody/contour/step_contour.py
```diff
@@ -0,0 +1,247 @@
...
+
+    Examples
+    --------
+    Calculate the global variation of a step contour
+    >>> contour = [60, 60, 60, 60, 62, 62, 62, 62, 63, 63, 63, 63, 63, 63, 63, 63, 63, 63,
+    ...           63, 63, 63, 63, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65, 65,
+    ...           72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72, 72,
+    ...           72, 72, 72, 72, 72, 72, 72, 72, 72, 72]
+    >>> step_contour_global_variation(contour)
+    4.463392767839281
```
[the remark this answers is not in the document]
pmcharrison: In fact doctest doesn't need rounding (as currently configured) so you should be able to just give a smaller number of dps

```
 >>> step_contour_global_variation(contour)
    4.463392
```

musmart/melody/contour/step_contour.py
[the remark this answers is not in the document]
pmcharrison: I don't think there can be a NaN in contour, can there?

amads/melody/contour/step_contour.py
```diff
@@ -0,0 +1,259 @@
...
+"""
+
+__author__ = "David Whyatt"
+
+import numpy as np
+
+
+class StepContour:
+    """Class for calculating and analyzing the step contour of a melody.
+    Rests are considered as extending the duration of the previous note.
```
pmcharrison, line 13: It would be good to add a brief overview of what step contour is here.

amads/melody/contour/step_contour.py
```diff
@@ -0,0 +1,259 @@
...
+            raise ValueError(
+                f"The length of pitches (currently {len(pitches)}) must be equal to "
+                f"the length of onsets (currently {len(onsets)})"
+            )
+
+        # Calculate durations from onsets
+        durations = []
+        for i in range(len(onsets) - 1):
+            durations.append(onsets[i + 1] - onsets[i])
+        # Add final duration (assume same as previous or 1.0 if first note)
```
pmcharrison, line 81: It's a strange heuristic isn't it -- why not just have the user implement the durations in advance? -- but presumably you're doing it this way because that's what Müllensiefen did?

Dav8Circle: In the original scripts, there's an inconsistency between different functions:
`step.contour <- function(pitch,dur16)` vs `line.contour <- function(onset, pitch)` 
(`line.contour` is used to calculate `int.contour` down the line)
Müllensiefen chooses to make a distinction between onsets and durations in the original (likely due to the units he defines for each one) but I feel that distinction is unhelpful for our use case - surely it would be better for a user to pass pitches and onsets into either/both functions and have them work?
Perhaps once it's clearer how we take rhythm information from the Score object, we can revise this?

pmcharrison: Just to understand, are you saying that Mullensiefen's original implementation of step contour takes durations as the input, but our implementation takes onsets? And you changed ours to onsets for consistency with line contour?

Dav8Circle: Yes, that's correct

pmcharrison: Ah, well I think we better avoid making the change then -- consistency is good, but I think this is too much of a sacrifice to make (guessing the duration of the last note).

I wonder if the best way to ensure a consistent interface is to make all these classes take a `Score` as their input, like @rbdannenberg was saying?

Or maybe we should be exposing helper functions like `interpolation_contour_variation` that take `Score` as their input? 

I'm not sure what's best, and to be honest the interface can easily be changed in the future, but I do think the important thing is that we don't sacrifice the accuracy of the algorithm by guessing durations

Dav8Circle: Yeah I can see that guessing the final duration is problematic - I've reverted back to durations as the input parameter and we can talk about interfacing another time

pmcharrison (approved):

pmcharrison:
@Dav8Circle feel free to merge if happy

Issue #29, opened 2024-12-05
Classes vs. functions for "models" with several possible "features"


HuwCheston:
Currently we seem to have adopted the principle of using functions for features that return individual values. 

This makes sense when we expect the user to only desire a single return value. For instance, `musmart.algorithms.complexity.lz77` provides a `lz77_size` function that takes a discrete sequence and returns the complexity as a numerical value. If the user wants to get the result from the underlying `lz77` algorithm, they can use the `lz77_compress` function.

The situation becomes a bit more complex when we have a single algorithm where the user might want numerous different return values. For instance, linear phase correction modelling of coordination (as in [this paper](https://doi.org/10.1525/mp.2024.42.1.48)) has multiple potential return objects: we can return the "coupling" from one performer to another (or one performer to themselves), and for each "coupling" instance we have both raw coefficients, confidence intervals, standard errors, etc.

Implementing multiple functions to provide each return value is possible, but redundant if the user wants multiple objects from the same inputs as we'd end up computing the (relatively costly) model multiple times. In this case, a class-based structure might be better, where we compute the model a single time and provide all features that the user may wish to access as attributes of the class. 

If this is the case, we'd need to decide on the API that these objects should take, e.g., whether they should inherit from a base class  that defines the expected structure (in `sklearn` there is a `BaseEstimator`). A function could also be implemented that simply returns the class instance if we want to keep the function-based nomenclature, for example:

```
class PhaseCorrectionModel:
    def __init__(self, *args):
        self.attribute = ...    # some attribute defined here


def phase_correction_model(*args) -> PhaseCorrectionModel:
    return PhaseCorrectionModel(*args)


model = phase_correction_model()
model.attribute    # or even model["attribute"] with suitable overriding of __getitem__
```

PR #40, opened 2024-12-09
Interpolation contour

dmwhyatt:
Implements the Interpolation Contour and associated features from FANTASTIC toolbox (Müllensiefen, 2009). This approach takes a list of times (float) and a list of pitches (int) and returns a list of the interpolated gradient between the contour extremes of the input sequence. Provides functions to calculate the global direction of the contour, abs mean, standard deviation, ratio of direction changes, and a string of 4 letters to summarize the contour at 4 equally spaced points.


pmcharrison:
Thanks @Dav8Circle! Happy to review in detail soon. First, though....

Can you look at the following discussion? https://github.com/music-computing/amads/issues/29. I think that the equivalent here would be something like

```py
res = get_interpolation_contour(...) 
res.global_direction
res.gradients_std_dev
```

etc.

Also, can you double-check the failing test?

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
+"""
+Calculates the Interpolation Contour of a melody, along with related features, as
+implemented in the FANTASTIC toolbox of Müllensiefen (2009).
```
pmcharrison, line 3: Does Müllensiefen provide any original citation for this?

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+"""
+
+__author__ = "David Whyatt"
+
+import numpy as np
+
+
+class InterpolationContour:
+    """Class for calculating and analyzing interpolation contours of melodies."""
```
pmcharrison, line 12: It would be wise to document the citation etc in the class too so that people have it at hand.

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+        Excludes changing notes (notae cambiatae).
+
+        Parameters
+        ----------
+        pitches : list[int]
+            List of MIDI pitch values
+
+        Returns
+        -------
+        numpy.ndarray
```
pmcharrison, line 59: Your return value is inconsistent with the type hint in the function signature

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+            prev_pitch = pitches[i - 1]
+            curr_pitch = pitches[i]
+            next_pitch = pitches[i + 1]
+
+            # Skip changing notes (where adjacent pitches are equal)
+            if prev_pitch == next_pitch:
+                continue
+
+            # Check if it's a peak or valley
+            if (prev_pitch < curr_pitch and next_pitch < curr_pitch) or (
```
pmcharrison, line 80: Better to write

```py
is_peak = prev_pitch < curr_pitch and next_pitch < curr_pitch
is_valley = ...
```

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+                prev_pitch > curr_pitch and next_pitch > curr_pitch
+            ):
+                # Additional check for equal adjacent notes
```
pmcharrison, line 83: Doesn't this fail if you have more than one equal adjacent note? Isn't a better way of dealing with equal adjacent notes to preprocess by removing them?

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+                    ) < 0
+                if i < n - 2 and next_pitch == curr_pitch:
+                    is_extremum = (prev_pitch - curr_pitch) * (
+                        pitches[i + 2] - curr_pitch
+                    ) < 0
+                if is_extremum:
+                    extrema_indices.append(i)
+
+        extrema_indices.append(n - 1)  # Add last note
+        return np.array(extrema_indices)
```
pmcharrison, line 97: How come we are returning a np.array instead of a normal list? I think we are meant to do the latter.

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # Find local extrema
+        for i in range(1, n - 1):
+            # Check previous and next notes
+            prev_pitch = pitches[i - 1]
+            curr_pitch = pitches[i]
+            next_pitch = pitches[i + 1]
+
+            # Skip changing notes (where adjacent pitches are equal)
+            if prev_pitch == next_pitch:
```
pmcharrison, line 76: You should definitely make sure your tests cover all these coverage warnings

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+                    is_extremum = (prev_pitch - curr_pitch) * (
+                        pitches[i + 2] - curr_pitch
+                    ) < 0
+                if is_extremum:
+                    extrema_indices.append(i)
+
+        extrema_indices.append(n - 1)  # Add last note
+        return np.array(extrema_indices)
+
+    def calculate_interpolation_contour(self) -> list[float]:
```
pmcharrison, line 99: I think it's redundant to have a method called `calculate_interpolation_contour`, I would just put this all into the `__init__` function? Up to you though

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+        extrema_pitches = [self.pitches[i] for i in extrema_indices]
+
+        # Step 3: Calculate gradients
+        gradients = np.diff(extrema_pitches) / np.diff(extrema_times)
+
+        # Step 4: Calculate durations
+        durations = np.diff(extrema_times)
+
+        # Step 5: Remove durations below 0.05 seconds
+        durations[durations < 0.05] = 0
```
pmcharrison, line 125: 0.05 s should be a parameter, not hard-coded

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+
+__author__ = "David Whyatt"
+
+import numpy as np
+
+
+class InterpolationContour:
+    """Class for calculating and analyzing interpolation contours of melodies."""
+
+    def __init__(self, times: list[float], pitches: list[int]):
```
pmcharrison, line 14: Try and be consistent across your modules in whether times comes first or pitches. I think I would probably do pitch first, then time?

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+        """Initialize with time and pitch values.
+
```
pmcharrison, line 16: Any particular reason why InterpolationContour takes onset times but StepContour takes durations?

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+        Returns
+        -------
+        list[int]
+            List containing the signs (+1, 0, -1) of the gradient values in the contour
+
+        Examples
+        --------
+        >>> ic = InterpolationContour([0, 0.1, 0.2, 0.3], [60, 58, 64, 65])
+        >>> ic.direction[:4]  # Show first 4 values
```
pmcharrison, line 146: This is a rather incomplete test that does not illustrate the full functionality!

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+        # Step 4: Calculate durations
+        durations = np.diff(extrema_times)
+
+        # Step 5: Remove durations below 0.05 seconds
+        durations[durations < 0.05] = 0
+
+        # Step 6: Create weighted gradients vector by repeating each gradient
+        # proportional to its duration
+        # Convert durations to integer number of samples (at 100Hz sampling rate)
+        samples_per_duration = np.round(durations * 100).astype(int)
```
pmcharrison, line 130: Probably 100 Hz should be a parameter too? That would help you with testing.

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+"""
+
+__author__ = "David Whyatt"
+
+import numpy as np
+
+
+class InterpolationContour:
+    """Class for calculating and analyzing interpolation contours of melodies."""
+
```
pmcharrison, line 13: I feel like we need much more documentation of what interpolation contour actually is doing.

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+        25.9272...
+        """
+        return float(np.std(self.contour))
+
+    @property
+    def direction_change_ratio(self) -> float:
+        """Calculate the ratio of direction changes in the interpolation contour.
+
+        This measures the number of changes in contour direction relative to the number
+        of interpolation lines (i.e. number of different gradient values).
```
pmcharrison, line 193: 'interpolation lines'? Seems confusing

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+
```
pmcharrison, line 194: Could be worth explaining a bit more what the intuition is behind this.

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+        -------
+        str
+            String of length 4 containing letters a-e representing the gradient
+            categories at 4 equally spaced points in the contour
+
+        Examples
+        --------
+        >>> ic = InterpolationContour([0, 0.1, 0.2, 0.3], [60, 58, 64, 65])
+        >>> ic.class_label
+        'aaee'
```
pmcharrison, line 244: Where did you get this example output from, are you sure it's right?

pmcharrison: Seems surprising that the final one would be strong upward

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+        return float(direction_changes / total_changes)
+
+    @property
+    def class_label(self) -> str:
+        """Classify an interpolation contour into gradient categories.
+
+        The contour is sampled at 4 equally spaced points and each gradient is
+        normalized and classified into one of 5 categories:
+
+        - 'a': Strong downward (-2) - normalized gradient <= -1.45
```
pmcharrison, line 228: You need to give the units of normalized gradient

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # Sample the contour at those indices
+        sampled_points = [self.contour[i] for i in indices]
+
+        # Normalize the gradients to a norm where value of 1 corresponds to a semitone
+        # change in pitch over 250ms. Given that base pitch and time units in
+        # FANTASTIC are 1 second and 1 semitone respectively, just divide by 4
+        norm_gradients = np.array(sampled_points) * 0.25
+
+        classes = []
```
pmcharrison, line 259: You can simplify slightly by writing

```py
classes = ""
classes += "a"
```

etc

pmcharrison:
Good draft!

amads/melody/contour/interpolation_contour.py
Dav8Circle: fixed

amads/melody/contour/interpolation_contour.py
Dav8Circle: fixed

amads/melody/contour/interpolation_contour.py
Dav8Circle: fixed

amads/melody/contour/interpolation_contour.py
Dav8Circle: I've updated the implementation after testing robustly against the original implementation - think I'd prefer to keep in a seperate method since it's ended up quite complex for inclusion in `__init__`

amads/melody/contour/interpolation_contour.py
Dav8Circle: In the more accurate ported version, this is handled by multiplying by 10 with a rounding - this is exactly how it's implemented in FANTASTIC, but i'm not opposed to extending the functionality per se

amads/melody/contour/interpolation_contour.py
Dav8Circle: I don't want to shirk responsibility, but that's how it's written in R! I agree that consistency would be good to have though

amads/melody/contour/interpolation_contour.py
Dav8Circle: As above, maybe we can improve on this implementation a bit

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,314 @@
...
+        extrema_indices.append(0)
+
+        # Find local extrema
+        for i in range(1, n - 1):
+            # Check previous and next notes
+            prev_pitch = pitches[i - 1]
+            curr_pitch = pitches[i]
+            next_pitch = pitches[i + 1]
+
+            # Skip changing notes (where adjacent pitches are equal)
```
pmcharrison, line 98: As discussed -- this changing note logic gives unintuitive results, e.g. contour of 0 for `C D E D C`. Propose making a parameter called `method` which could take values of `"amads"`, `"fantastic"`, and then for our method we skip this changing note elimination logic.

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,314 @@
...
+            if prev_pitch == next_pitch:
+                continue
+
+            # Check if it's a peak or valley
+            is_peak = prev_pitch < curr_pitch and next_pitch < curr_pitch
```
pmcharrison, line 103: This peak detection logic would best be done by first removing duplicate adjacent notes, and then running your code

```py
            is_peak = prev_pitch < curr_pitch and next_pitch < curr_pitch
            is_valley = prev_pitch > curr_pitch and next_pitch > curr_pitch
```

and then skipping the "Additional check for equal adjacent notes".

Proposal we agreed on: support both behaviours, but choose which one depending on the `method` parameter.

amads/melody/contour/interpolation_contour.py
Dav8Circle: updated!

amads/melody/contour/interpolation_contour.py
Dav8Circle: updated with `method=` as discussed, with new, intuitive doctests

amads/melody/contour/interpolation_contour.py
Dav8Circle: added

amads/melody/contour/interpolation_contour.py
Dav8Circle: fixed!

amads/melody/contour/interpolation_contour.py
Dav8Circle: implemented

amads/melody/contour/interpolation_contour.py
Dav8Circle: implemented

amads/melody/contour/interpolation_contour.py
Dav8Circle: I regret that Müllensiefen doesn't actually discuss why this might be useful in the tech report - I suppose it tells us something about global direction variability?

amads/melody/contour/interpolation_contour.py
pmcharrison: Still todo -- I don't think you need to write much necessarily, just a sentence or two explaining what interpolation contour is

amads/melody/contour/interpolation_contour.py
pmcharrison: Perhaps you could write something like "The proportion of contour changes (e.g. from -1.2 to -0.8, or from 1 to 0.5) that constitute changes in direction (e.g. from -0.5 to 0.25)." Did I get that right?

pmcharrison:
Thanks @Dav8Circle, making good progress. For both this and the other PR it would be good idea to go through carefully and think where some more documentation could help the user understand what the algorithms are computing for them.

amads/melody/contour/interpolation_contour.py
pmcharrison: I think better to write

> expressed in units of semitones per 0.25 seconds

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,407 @@
...
+        return [float(x) for x in interpolation_contour]
+
+    @staticmethod
+    def _calculate_amads_contour(pitches: list[int], times: list[float]) -> list[float]:
+        reversals_pitches = []
+        reversals_time = []
+        reversals_pitches.append(pitches[0])
+        reversals_time.append(times[0])
+
+        # Handle repeated notes - keep only middle note when 3 consecutive notes are equal
```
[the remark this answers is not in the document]
Dav8Circle: Yeah that was incomplete logic, my bad! I've updated the code such that if there are repeated notes, it keeps only the **middle** values, which I think is better in-keeping with the notion of interpolation contour (rather than a front or back loaded gradient line). Then the reversals are identified and kept for interpolation, as they should've been all along!

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,407 @@
...
+        1.0
+
+        FANTASTIC method returns 0.0 for this example
+        >>> ic = InterpolationContour([60, 62, 64, 62, 60], [0, 1, 2, 3, 4], method="fantastic")
+        >>> ic.direction_changes
+        0.0
+        """
+        # Convert contour to numpy array for element-wise multiplication
+        contour_array = np.array(self.contour)
+        print(contour_array)
```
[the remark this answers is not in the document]
Dav8Circle: oops, well spotted, done!

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,407 @@
...
+            Ratio of the number of changes in contour direction relative to the number
+            of different interpolated gradient values
+
+        Examples
+        --------
+        >>> ic = InterpolationContour([60, 62, 64, 62, 60], [0, 1, 2, 3, 4])
+        >>> ic.direction_changes
+        1.0
+
+        FANTASTIC method returns 0.0 for this example
```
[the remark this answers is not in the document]
Dav8Circle: I think the problem is the way that the turning points trigger subsequent logic: there's a code block that's triggered by `length(candidate.points.pitch) > 2` which enables a lot of the logic, including the addition of the final note to the turning point vector. 

If one passes a melody into this function that only has one turning point, the length of `candidate.points.pitch` would be exactly 2, insufficient to progress to casting these candidates to `turning.points`. The result of a melody with only one turning point ends up being exactly the gradient between the first and last note in the melody.

I think that's what's going on, but I'm not very good at debugging R yet!

amads/melody/contour/interpolation_contour.py
```diff
@@ -0,0 +1,407 @@
...
+        """Initialize with pitch and time values.
+
+        Parameters
+        ----------
+        pitches : list[int]
+            Array of pitch values
+        times : list[float]
+            Array of onset times in seconds
+        method : str, optional
+            Method to use for contour calculation, either "fantastic" or "amads".
```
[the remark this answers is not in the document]
Dav8Circle: done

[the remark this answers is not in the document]
rbdannenberg: Why use pitches and times  as parameters when we have a score representation that includes this information and is the representation obtained from reading musicxml and MIDI?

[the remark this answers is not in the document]
pmcharrison: I think we agreed earlier on that all the algorithms should be invokable with minimal inputs (e.g. lists of floats), but we should certainly support Scores as inputs though. One possibility is something like this:

```py
class InterpolationContour:
    @classmethod
    def from_score(score: Score, **kwargs):
        notes = score.flatten().find_all(Note)
        pitches = [n.pitch.keynum for n in notes]
        onsets = [n.start for n in notes]
        return InterpolationContour(pitches, onsets, **kwargs)
```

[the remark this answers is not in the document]
rbdannenberg: Yes, I'd be even happier to see more uniform use of scores and perhaps have a simple score constructor from lists of pitches and durations. Where are these pitches and durations going to come from anyway? Is that an important use case? On the other hand, basics.py (Score) should have a standard method for `flatten().find_all(Note)` since this is common. There are some details to work out: do we want a list or an interator, if an iterator, how do we access the length? (some algorithms need it at the beginning)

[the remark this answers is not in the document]
pmcharrison: +1 for your idea of simple score constructor from lists of pitches and durations! For example something like 

```py
score = Score.from_melody(pitches=List, onsets=Optional[List], durations=Optional[List]) 
```

Then agreed, would be great to have something like `score.get_notes()`, or perhaps `score.notes`. The latter is neater but I suppose hides the fact that a non-trivial operation is happening underneath.

I would be open to `score.notes` being either a generator or a list. The generator is more flexible in principle but it means people have to write `len(list(score.notes))` which is a bit ugly. I suppose you could have `score.iter_notes()` being a generator and `score.notes` being a list, but maybe that is getting unnecessarily complex.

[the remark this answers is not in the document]
pmcharrison: See #57 and #58 for draft implementations of `get_notes` and `from_melody`. I propose that we merge this PR in the meantime to get the code onto main, but we create a subsequent issue for revising the interface such that the top-level class accepts a `Score` as the input?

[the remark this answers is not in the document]
pmcharrison: I think the discussion happened elsewhere, but I think I have been persuaded that it makes sense for all top-level functions to take Scores as input.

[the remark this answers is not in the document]
Dav8Circle: I'm happy with this!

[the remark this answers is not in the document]
pmcharrison: Thanks! can you make an issue?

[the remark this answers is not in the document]
Dav8Circle: Done!

pmcharrison (approved):
@Dav8Circle documentation looks great, many thanks. I'll approve! Please feel free to merge it if you're happy.

pmcharrison:
PS I would recommend the 'squash and merge' option.

PR #48, opened 2024-12-17
Meter

MarkGotham:
Start work on existing metrical functionality.

More to be do (splitting files, other algorithms etc.) but this a stable start.


pmcharrison:
Thanks @MarkGotham! Will look soon. In the meantime, did you see the error in the tests?

```
Pre-commit checks failed! If you are seeing this message, you probably forgot to install the pre-commit hooks on your local machine. To fix this, run the following in your terminal:

pip install pre-commit && pre-commit install

Next time you commit, pre-commit will run automatically, and will fix or flag any issues.
To see all issues without committing, you can run:

pre-commit run --all-files
```

MarkGotham:
I'll keep pecking away at this as time allows. 

Btw @pmcharrison  the py3.9 check seems to be hanging

pmcharrison:
Hi @MarkGotham, it looks like a very useful contribution! There is quite a lot of code to go through and it interacts with some questions we have in our ongoing email chain about relationships to other encoding schemes. Is it OK if I let that discussion develop some more before going through the code in detail? Let me know if you're anxious to merge the code sooner rather than later. Hanging onto the merge request for now has some benefits (e.g. the line-by-line code coverage and the easy commenting) which we would lose if we merge early.

amads/time/meter/break_it_up.py
```diff
@@ -365,9 +484,8 @@ def start_hierarchy_from_ts(ts_str: str, minimum_pulse: int = 64):
 
 
 def starts_from_ts_and_levels(
-    ts_str: str,  # TODO Union[str, meter.time_signature] or similar
+    ts_str: str,
```
pmcharrison, line 487: I think we had agreed to use full names rather than abbreviations in Python expressions. Do you think that is impractical here?

rbdannenberg: It's at least borderline for me. Using some conventions like ts for time signature and dur for duration seems reasonable. Philosophically, there's a reason we use abbreviations, even  in formal writing, etc. :-)

pmcharrison: I think often there are ways around it, though. For example here I think the problem is that we are trying to code the input type into the function name. This would be a natural application for OOP, I think:

```py
class TimeSignature:
    def get_starts(levels=None):
        ...
```

I think now that IDEs have autocomplete there is little overhead for using full words, and it does generally enhance readability. `dur` is normally quite obvious from context but I wasn't sure what `ts` meant until I read the documentation. This matters particularly in a package when we are taking a liberal approach to namespace exporting and so the context is broad.

MarkGotham: Yes that's all fine:
- no rush with the review.
- continue discussion via email.
- probably full names at least for anything not immediately obvious.

One objection (possibly trivial): the notion of a `TimeSignature` class. As I see it, time signatures are either trivial (just two numbers?), or basic/under-defined metrical structures, or else not in scope here. Here as elsewhere, I find the maxim "we're not in the business of making notation" a helpful line in the sand.

pmcharrison: Yes, in this case (having now looked in the code) this is actually the only use of time signature, more or less, so creating a time signature class would be overkill. I think this also means, though, that abbreviating time signature is a bad idea because it would be an abbreviation that only gets used for one function

MarkGotham:
- Done: basic re-name `_time_signature` (and some other bits)
- TODO: integrate conversion routines into OOP

pmcharrison:
> * Done: basic re-name `_time_signature` (and some other bits)
> * TODO: integrate conversion routines into OOP

Cool! Maybe don't worry about OOP refactoring at this point, I think my specific suggestion about TimeSignature might have been out-of-place here, maybe we can brainstorm organisation strategies at some point together once I've gone through it properly

pmcharrison:
To be honest at the outset an expedient strategy could be to prioritise polishing the public API (exported functions) and the private side can be finessed at our leisure

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    """
+    Create a start hierarchy for almost any time signature
+    directly from a string (e.g., "4/4") without dependencies.
+
+
+    Parameters
+    ----------
+    ts_str
+        Any valid string repersenting a time signature. See examples below.
+    enforce_2s_3s
```
pmcharrison, line 180: Perhaps a more general version of `enforce_2s_3s` would be instead `permitted_subdivisions=[2, 3]` or something like that? Seems like an easy way to be more general, though I appreciate this flexibility would not often be used....

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+
+
+# ------------------------------------------------------------------------------
+
+
+def starts_from_time_signature(
+    ts_str: str, enforce_2s_3s: bool = True, minimum_pulse: int = 64
+) -> list:
+    """
+    Create a start hierarchy for almost any time signature
```
pmcharrison, line 172: It would be good to explain the underlying heuristics being used here, for example to determine whether 6/4 should be grouped into 3s or 2s. If the heuristics only work for a certain set of time signatures, it would be worth stating that explicitly.

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    """
+
+    # Prep and checks
+    numerator, denominator = ts_str.split("/")
+    numerators = [
+        int(x) for x in numerator.split("+")
+    ]  # Note: support "+" before one "/", e.g., `2+3/4`
+    denominator = int(denominator)
+
+    supported_denominators = [1, 2, 4, 8, 16, 32, 64]
```
pmcharrison, line 243: Is there a rationale for encoding 64 as the maximum value here? Can't we programmatically respond to greater values of minimum_pulse?

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+            f" {supported_denominators}."
+        )
+    if minimum_pulse not in supported_denominators:
+        raise ValueError(
+            "Invalid minimum_pulse: it should be expressed as a note value, from one of"
+            f" {supported_denominators}."
+        )
+
+    # Prep. "hiddenLayer"/s
+    hidden_layer_mappings = (
```
pmcharrison, line 256: Can this be written more clearly? Perhaps it would suffice to comment what this data structure is and what it's being used for.

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+
+    """
+    if levels is None:
+        levels = range(6)
+
+    if max(levels) > 6:
+        raise ValueError("6 is the maximum level depth supported.")
+    levels = sorted(levels)  # smallest first
+    if 0 not in levels:  # likely, and not a problem
+        levels = [0] + levels  # Always have level 0 as first entry
```
pmcharrison, line 365: If you refactor to implement `MetricalHierarchy.from_time_signature` as suggested above, then I think you could implement this quite succinctly as:

```py
    @classmethod
    def from_time_signature(cls, time_signature: str, levels: Optional[List[int]]):
        numerator, denominator = time_signature.split("/")
        ...
        hierarchy = ...
        levels = set([0] + levels)
        hierarchy = [pulses for level, pulses in hierarchy if depth in levels]
        return cls(hierarchy)
        ....
```

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+        this_ql = 4 / this_denominator
+        starts = starts_from_lengths(measure_length, this_ql)
+        start_hierarchy.append(starts)
+
+    return start_hierarchy
+
+
+def starts_from_time_signature_and_levels(
+    ts_str: str,
+    levels: Union[list, None] = None,
```
pmcharrison, line 309: Are you sure you want `levels` to be a list of all levels to be included? so that users can skip consecutive levels? might it be more natural to rename this `depth` and make this be the total number of levels to include?

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    return starts_by_level
+
+
+def starts_from_pulse_lengths(
+    pulse_lengths: list,
+    measure_length: Optional[float] = None,
+    require_2_or_3_between_levels: bool = False,
+):
+    """
+    Convert a list of pulse lengths into a corresponding list of lists
```
pmcharrison, line 382: Clarify that the user is providing the pulse lengths for each level of the metrical hierarchy, and the algorithm is expanding this into a hierarchy assuming isochrony.

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+
+    >>> qsl[2]
+    [0.0, 1.0, 2.0, 3.0, 4.0]
+
+    >>> qsl[3]
+    [0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0]
+
+    """
+
+    pulse_lengths = sorted(pulse_lengths)[::-1]  # largest number first
```
pmcharrison, line 439: This is cleaner (in-place sorting):

```
pulse_lengths.sort(reverse=True)
```

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    return start_list
+
+
+def coincident_pulse_list(
+    m: MetricalHierarchy,
+    granular_pulse: float,
+) -> list:
+    """
+    Create a flat list in the form [4, 1, 2, 1, 3, 1, 2, 1]
+    from the number of intersecting pulses at each position.
```
pmcharrison, line 473: This definition could be made clearer

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    These are listed in order of override, e.g., if a full `start_hierarchy` is specified, all else are ignored.
+
+    For a 'quick start', the `time_signature` defaults will serve most use cases, and
+    that along with `levels` should be cover almost all.
+
+    Parameters
+    ----------
+    start_hierarchy
+        Users can specify the `start_hierarchy` directly
+        and completely from scratch (ignoring all other parameters and defaults).
```
pmcharrison, line 39: It would be good to include a more formal definition of start_hierarchy here. I'm not entirely sure about whether it's good to include 4.0 as the final element in 4/4, for example, because in my mind 4.0 is equal to 0.0 in the relevant modular arithmetic

pmcharrison: An alternative would be to encode bar length (or cycle length??) as a separate attribute of `MetricalHierarchy`

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    At least one of the parameters (below) must be specified.
+    These are listed in order of override, e.g., if a full `start_hierarchy` is specified, all else are ignored.
+
+    For a 'quick start', the `time_signature` defaults will serve most use cases, and
+    that along with `levels` should be cover almost all.
+
+    Parameters
+    ----------
+    start_hierarchy
+        Users can specify the `start_hierarchy` directly
```
pmcharrison, line 38: Are we assuming that the highest level of the hierarchy is the measure? Presumably it would be nice to capture hypermetre too?

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+        and completely from scratch (ignoring all other parameters and defaults).
+        Use this for advanced, non-standard metrical structures
+        including those without 2-/3- grouping, or even nested hierarchies.
```
pmcharrison, line 41: Should we state explicitly (and test that) all elements in a given level of the hierarchy should be present in the lower levels of the hierarchy too? In other words, we are assuming that our metrical hierarchy is created by recursively subdividing a unit of time? Maybe worth explicitly stating this, and acknowledging that this means we can't encode polymetres.

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    >>> coincident_pulse_list(m, granular_pulse=0.5)
+    [3, 0, 1, 0, 2, 0, 1, 0]
+
+    """
+    if isinstance(m, MetricalHierarchy):
+        m = m.start_hierarchy
+    elif not isinstance(m, list):
+        raise ValueError("The X must be a `MetricalHierarchy` object or a list")
+
+    measure_length = m[0][-1]
```
pmcharrison, line 503: Are we sure this is a good assumption to make? see above

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    start_list = []
+
+    for pulse_length in pulse_lengths:
+        starts = starts_from_lengths(measure_length, pulse_length)
+        start_list.append(starts)
+
+    return start_list
+
+
+def coincident_pulse_list(
```
pmcharrison, line 467: I'm a bit nervous that this function relies on floating point equality. Better to implement some kind of tolerance for approximate matching?

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    --------
+
+    >>> starts_from_lengths(4, 1)
+    [0.0, 1.0, 2.0, 3.0, 4.0]
+
+    >>> starts_from_lengths(4, 1, include_measure_length=False)
+    [0.0, 1.0, 2.0, 3.0]
+    """
+    starts = []
+    count = 0
```
pmcharrison, line 559: I'm not sure about including the rounding code in here, wouldn't it be better to implement a tolerance in comparison checks instead?

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    while count < measure_length:
+        starts.append(round(float(count), 4))
+        count += pulse_length
+
+    if include_measure_length:
```
pmcharrison, line 564: Maybe a more elegant version of this would be

```
    starts = [round(float(i), 4) for i in range(0, measure_length, pulse_length)]
    if include_measure_length:
        starts.append(round(float(measure_length), 4))
    return starts
```

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+
+    return starts
+
+
+def starts_from_beat_pattern(
+    beat_list: list, denominator: float, include_measure_length: bool = True
+):
+    """
+    Converts a list of beats
+    like [2, 2, 2]
```
pmcharrison, line 575: It's quite difficult to understand from this documentation what a list of beats is, or a list of starts

amads/time/meter/break_it_up.py
```diff
@@ -1,218 +1,177 @@
...
-    The input parameters are:
-
-    note_start_start:
-    the note or rest in question"s starting start.
-
-    note_length:
-    the note or rest"s length.
+        Split up notes and rest to reflect a specified metrical hierarchies.
+
+        This class
```
pmcharrison, line 24: Can you articulate a bit more clearly what this class is doing? What does it mean to 'split up notes and rests', and what are 'note-duration' pairs? Maybe it would help if you gave a more formal definition of the algorithm?

I think in general quite a few of the docstrings in this PR could be revisited with a view to making the functionality more transparent to new users, who won't necessarily want to read the source code to know how the algorithm works.

amads/time/meter/break_it_up.py
```diff
@@ -222,22 +181,22 @@ def level_pass(self):
...
+        find the next higher value from those levels.
         Used for determining iterative divisions.
         """
         for p in positions_list:
             if p > self.updated_start:
                 duration_to_next_position = p - self.updated_start
                 if self.remaining_length <= duration_to_next_position:
                     self.start_duration_pairs.append(
-                        (self.updated_start, self.remaining_length)
+                        (self.updated_start, round(self.remaining_length, 4))
```
pmcharrison, line 192: Are we sure rounding is safe? Might checking with a tolerance be better?

pmcharrison (commented):
Hi @MarkGotham, thanks for these contributions! I have left some initial comments, hope they provide some useful food for thought? Will be happy to discuss more.

amads/algorithms/__init__.py
```diff
@@ -19,7 +19,7 @@
 from ..polyphony.skyline import skyline
 from ..time.durdist1 import duration_distribution_1
 from ..time.durdist2 import duration_distribution_2
-from ..time.meter.break_it_up import ReGrouper
```
pmcharrison, line 22: Perhaps as part of this PR it could be useful to implement an 'example' notebook that illustrates how the functions can be applied for useful analysis approaches?

pmcharrison:
FYI @MarkGotham, I'm tagging @Dav8Circle here because he is ready to implement some melody analysis algorithms that would build on this MetricalHierarchy representation, things like melodic accent models, syncopation models etc. Once we have merged the PR he can get going with that

MarkGotham:
> FYI @MarkGotham, I'm tagging @Dav8Circle here because he is ready to implement some melody analysis algorithms that would build on this MetricalHierarchy representation, things like melodic accent models, syncopation models etc. Once we have merged the PR he can get going with that

Great! That's a nice reason to have to finish up ;)

amads/algorithms/__init__.py
MarkGotham: That would be useful ... but perhaps a task for a later PR given David's waiting for this?

pmcharrison: Yes agreed! Maybe even David provides one as part of his work

MarkGotham: By all means!

MarkGotham:
Ok, doing this now.

I'll focus on completing this PR in a credible form, so
- implementing and "resolving" agreed changes and
- moving any ongoing questions to an new issue.

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+
+    At least one of the parameters (below) must be specified.
+    These are listed in order of override, e.g., if a full `start_hierarchy` is specified, all else are ignored.
+
+    For a 'quick start', the `time_signature` defaults will serve most use cases, and
+    that along with `levels` should be cover almost all.
+
+    Parameters
+    ----------
+    start_hierarchy
```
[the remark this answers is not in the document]
MarkGotham: I'm open to options. As it stands, I'm aiming to use `start` in a way that's consistent with the wider repo and distinguish `hierarchy` where directly stated from cases like `pulse_lengths` where it's not (albeit implied). Currently `start_hierarchy` is the only explicit hierarchy among the entry options, though that will likely change (expand to other kinds).

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    of metrical levels in terms of starts expressed by quarter length
+    from the start of the measure,
+    and (optionally) other representations.
+
+    At least one of the parameters (below) must be specified.
+    These are listed in order of override, e.g., if a full `start_hierarchy` is specified, all else are ignored.
+
+    For a 'quick start', the `time_signature` defaults will serve most use cases, and
+    that along with `levels` should be cover almost all.
+
```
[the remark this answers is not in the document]
MarkGotham: OK, duly done, though the doc comments and examples appear together right? So you;d hope anyone curious would read on ...

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+        Users can specify the `start_hierarchy` directly
+        and completely from scratch (ignoring all other parameters and defaults).
+        Use this for advanced, non-standard metrical structures
+        including those without 2-/3- grouping, or even nested hierarchies.
+
+    pulse_lengths
+        In absense of a `start_hierarchy`, users can specify the structure by
+        `pulse_lengths` e.g., in the form [4, 2, 1].
+        This will be converted into a `start_hierarchy`.
+        See `starts_from_pulse_lengths` for further explanation.
```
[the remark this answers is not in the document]
MarkGotham: Disagree on this point: the `starts_from_pulse_lengths` function sorts the list before being used, so the ordering is handled internally. E.g., user could enter `[1, 4, 2]` and get the same results.

amads/time/meter/__init__.py
```diff
@@ -0,0 +1,619 @@
...
+    """
+
+    def __init__(
+        self,
+        time_signature: Optional[str] = None,
+        levels: Optional[list] = None,
+        pulse_lengths: Optional[list] = None,
+        start_hierarchy: Optional[list] = None,
+        names: Optional[dict] = None,
+    ):
```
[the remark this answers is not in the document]
MarkGotham: I'm open to this though: 1) as you say, we can consider this refactor later and 2) I tend to prefer keeping static functions out of the class to support use cases that only require the function, leaving the class to handle cases where you might need to navigate multiple representations and/or make _changes_ to a metrical structure in place (rare, I would think).

For now, I've made a `get_starts` method to provide some clean up, short of the wholesale change.

Tagging @rbdannenberg in case you've a strong feeling on this core design question. Issue to follow

amads/time/meter/__init__.py
MarkGotham: That's certainly an option. Would you constrain that list? Moving this one to the issue.

MarkGotham: That's certainly an option. Would you constrain that list? Moving this one to the issue.

amads/time/meter/__init__.py
MarkGotham: Ok. Again, I think the clearest explanation is in the `Examples`. I'll add a comments to this effect and re-work them a little for clarity of this specific point.

amads/time/meter/__init__.py
MarkGotham: Programmatically, we could indeed respond to greater values, but there is psychological evidence for structures beyond this ceasing to be 'meter' in the same sense. While there was once a practice of extending 'metrical structure' into the realm of 'form' (e.g., Cooper and Meyer), that seems to be a poor reflection of the psychology of meter (see e.g., London _Zeitnetz_).

I consider that reason enough and note that this is more principled than other restrictions: we're also limited to single metres (not mixtures that would make a case for 96 divisions of the bar, for instance), and to exclude the inaptly named 'irrational' metres (e.g., `X/6`). 

I've added a comment on the docs, but a shorter one than this ;)

amads/time/meter/__init__.py
MarkGotham: Yup good shout. 

I'll also move it out of the function.

Actually, come to think of it, that's another answer to the debate above. The user could create these mappings. I'll implement that, that's probably enough for a separate commit ...

amads/time/meter/__init__.py
MarkGotham: Both the refactor and the use of `depth` are open questions moving to the issue.

amads/time/meter/__init__.py
MarkGotham: I agree that depth is more common, but I prefer the more flexible option in this case.

While the combination of  `4` and `1` tends to be taken to imply an intervening `2`, I see a value in being able to code and work with the former (without the 'fill in') as a metrical structure in itself. This connects with general considerations like the usage of levels in practice, and specific cases like the distinction between 4/4 and 2/2 and the presence or absence of a half-bar division there.

amads/time/meter/__init__.py
MarkGotham: Yes we must and do capture hypermeter. 

When you make a metrical structure with an overall span of 4, that could be a single measure of 4/4 with no hypermeter, but it could also be in 2/4 with a layer of hypermeter.

It is core to this module that we make no such distinction.

Only in the case of building it from a time signature is that specified (by the user).

amads/time/meter/__init__.py
MarkGotham: Yes, let's have a `self.cycle_length = self.start_hierarchy[0][-1]` on the init. I'll give that its own commit.

Whether to include the full cycle value is open to discussion. Some routines require it but can be handled otherwise as you suggest.

amads/time/meter/__init__.py
MarkGotham: So this is consistent with current set up. If we refactor as above then this must change too. Heading to the issue. Only change for now is adding this as an attribute on the class.

amads/time/meter/__init__.py
MarkGotham: One to discuss - @rbdannenberg has raised this question for wider repo mattes. Issue.

amads/time/meter/__init__.py
MarkGotham: As above. Issue. @rbdannenberg

amads/time/meter/__init__.py
MarkGotham: Yes, as we've discussed offline. I have this on my TODO list for the testing side. Even in cases like 4/4 +3/8, any "levelled out" representation is going to have to have this property. Issue. For now, the comment only.

amads/time/meter/__init__.py
MarkGotham: This problem with this is that neither `measure_length` nor `pulse_length` are guaranteed to be integers.

MarkGotham: Adding an example to that effect.

amads/time/meter/break_it_up.py
MarkGotham: As above.

amads/time/meter/break_it_up.py
MarkGotham: Fair enough, with the caveat that the best explanations are in the documentation Examples as discussed in many points above.

pmcharrison: When you resolved this comment, was that indicating that you have made the changes?

If it helps, I think I would have done it something like this:

```py
class MetricalHierarchy;
    # Why not make it a method of `MetricalHierarchy`? 
    def split_event(
        start,  # If the input can be a rest, better to not call this argument `note_start`?
        duration,
        split_same_level = True
        """
        Splits the timespan of a note or rest into smaller timespans, following Western music notation conventions.

        In Western music notation it is conventional to break down the timespan of an event (e.g. note, rest)
        to facilitate counting out the event's duration. For example, suppose we are in 4/4 and we wish 
        to notate a note that begins one eighth note after the start of the bar, and that has a total duration 
        of three eighth notes. In this case we would split that note into two tied notes, the first being an 
        eighth note, the second being a quarter note. This allows the second note to start on the quarter note beat.

        Specifically, the procedure can be stated as follows. Identify the metrical level that the original event
        started at. As we progress through the duration of the event, if we encounter a beat at a higher metrical
        level than the one we started at, then terminate the event and create a new tied event at that time point.
        Identify the metrical level of this new tied note, and repeat the process until the end of the original event
        is reached.

        ...
```

I think it's valuable to try and convey the general high-level point before getting into the weeds of e.g. input types etc. Input types should normally be saved for the Parameters part of the docstring. I think Examples are valuable for illustrating the specifics of what objects go in and out of the routine, but I don't think they should be relied on for explaining the high-level algorithm.

amads/time/meter/__init__.py
pmcharrison: OK, good point, but easily solved by using `numpy.arange` instead of the built-in `range`?

amads/time/meter/__init__.py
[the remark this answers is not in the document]
MarkGotham: OK, I'm adding a pair of convenience methods that (I think)
- serve the alternative design you prefer (as above) ...
- ... without breaking the existing routes through
- ... and without duplicating code (everything is sent through the standalone).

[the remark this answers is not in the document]
pmcharrison: Cool!

I see code like this:

```
    def from_time_signature(
        self,
        time_signature: str = "2+2+2+3/8",
        enforce_2s_3s: bool = True,
        minimum_pulse: int = 64,
    ) -> None:
  
        self.time_signature = time_signature
        self.start_hierarchy = starts_from_time_signature(
            time_signature, enforce_2s_3s=enforce_2s_3s, minimum_pulse=minimum_pulse
        )
```

which makes me a bit nervous because I think in general we should have the class be immutable, so we don't instantiate it as an empty object and then make changes to core attributes outside the `__init__` function. 

I also think you're duplicating logic now, `starts_from_time_signature` is now called both in `from_time_signature` and `get_starts`, since the code takes different paths depending on whether you do `__init__` with `time_signature` or instead `from_time_signature`. I think if you want to implement these methods I suggested, you need to consolidate and only have one pathway.

The best pattern here, I think (what I think I recommended above) is that `from_time_signature` does some preprocessing of the arguments and then calls the original constructor function with those preprocessed arguments. 

Alternatively, you can flip the logic and have __init__ check whether `time_signature` is provided, and if so delegate to `cls.from_time_signature`. That seems more complicated though because you end up going `__init__ -> from_time_signature -> __init__` rather than just `from_time_signature -> __init__`.

pmcharrison:
Hi @MarkGotham, thanks for the chat yesterday! Here are some discussions of factory methods that illustrate more what I was talking about in my initial code review:

- https://www.attrs.org/en/stable/init.html
- https://stackoverflow.com/questions/682504/what-is-a-clean-pythonic-way-to-implement-multiple-constructors

The general principle is to keep `__init__`  minimal and try and split more complex construction logic into factory methods, which becomes the primary interface for users. This is still my preferred route, I reckon, but I think it would be ok to have a more complex `__init__` still as long as we manage the complexity well with helper functions. Maybe when we get our pair programming session we can try out the different options and see what feels cleanest.

MarkGotham:
Thanks for this @pmcharrison and to @rbdannenberg for a separate conversation that ended up being relevant.

I've been mulling this and have a new(?) proposal (with apologies if this is actually one of the previous suggestions @pmcharrison). None of the underlying implementations here would change (not much at least), only how they're packaged in objects.

Situation:
- We’ve moved pretty decisively away from the notion of simple data structures in favour of objects etc (all of us, throughout the project) so I can stop holding on to that as a design principle ;)
- This metrical functionality should offer multiple representations, but most users will only want to use one or two of them.
- There are so many types of representation that prospects for confusion are high, particularly in terms of keeping track of what is currently represented and checking that they align as expected.

As such, I’m tempted to swing quite far the other way. 
- Make separate classes for _everything_. E.g., classes for `StartTimeHierarchy` and `PulseLevels` etc.. 
- Each of these has several `.from` methods.  E.g., `StartTimeHierarchy.from_pulse_levels(1,2,4)`. 
- Make the _combination_ of these metrical hierarchy objects it's own class (e.g., `CombinedMeterRepresentations`), requiring these other classes to have been handled first.
    - Always run checks that the constituent parts are compatible when new attributes are added.
    - Pipe the `from` functionality here. E.g., `.derive_start_time_hierarchy` when there's a `pulse_values` attribute is the same functionality as `StartTimeHierarchy.from_pulse_levels()` (above). The difference is that we're clear on what (combination) is currently stored.
- This extends to the representations of weighting values I mentioned. They become simply another class to be included or derived (or neither).

An additional possibility (separate from the above) is to use numpy arrays throughout:
- This would make comparisons and transformations much easier. For example, establishing the _equivalence_ (NB: not _identity_) between metrical hierarchies. For instance, those based on pulse levels 1,2,4 and 2,4,8 are equivalent (in the absence of a tempo marking).
- This is even if we take the view that these hierarchies are immutable (certainly a reasonable position if we are making scores that way). We will likely still want to make a new metrical hierarchy based on another one, both for comparison and otherwise. For example, in the example case above 1,2,4 and 2,4,8 can be compared with arrays as simply  as `/=2`.
- Alternatively, the values are stored as tuples (NB: not lists) and then made into arrays briefly for that functionality (e.g., comparison).
- These arrays also potentially provides the basis for analysis as distinct from theory: to represent not only metrical structures in the abstract, but as the basis of a piece. We use several metrical objects which can be represented with or without their start time (this being simply numpy array +/- start time). A simple analysis could be in the form of (say) "32 times a metrical structure based on pulse levels, 1, 2, 4, 8, 16". Then a more detailed analysis could include weighting the relative presence of levels in each of those 32 iterations.

pmcharrison:
Hi @MarkGotham -- I am feeling positive about your new direction for implementation. Pros: I think that, once a data structure becomes more complex than a flat list or dict, it indeed becomes useful to represent it in a class. The resulting class ends up being easier to document and it organises related functionality in a coherent namespace. Also, it becomes easier to make subsequent changes to the internal representation of the class without breaking code that uses it.

Regarding this:

> Each of these has several .from methods. E.g., StartTimeHierarchy.from_pulse_levels(1,2,4).

I think you could then simplify it even more as

```py
pulse_levels = PulseLevels(...)
start_time_hierarchy = pulse_levels.to_start_time_hierarchy()  # no args required
```

i.e. `from_*` methods (with arguments) become `to_*` arguments (without arguments).

pmcharrison:
Hi @MarkGotham -- I am feeling positive about your new direction for implementation. Pros: I think that, once a data structure becomes more complex than a flat list or dict, it indeed becomes useful to represent it in a class. The resulting class ends up being easier to document and it organises related functionality in a coherent namespace. Also, it becomes easier to make subsequent changes to the internal representation of the class without breaking code that uses it.

Regarding this:

> Each of these has several .from methods. E.g., StartTimeHierarchy.from_pulse_levels(1,2,4).

I think you could then simplify it even more as

```py
pulse_levels = PulseLevels(...)
start_time_hierarchy = pulse_levels.to_start_time_hierarchy()  # no args required
```

i.e. `from_*` methods (with arguments) become `to_*` arguments (without arguments).

rbdannenberg:
I feel that I need to better understand the underlying motivation, e.g. theory and applications, of these representations. All the statements make sense, but a good representation should facilitate computation, and I'm not sure what the computation is. Are there papers I should read or plans for some algorithms using these representations? Also a big issue is (possibly) connecting descriptive or analysis data to actual Scores. Peter and I had some discussions of whether per-note attributes are kept in "parallel" arrays, so the ith element of some vector pertains to the ith Note of a Score, or do we simply set an attribute directly in each Note object (I like the latter.) Is there a similar argument for where does metrical hierarchy information go?

pmcharrison:
@rbdannenberg's question about how to represent the annotations is an interesting one. I think it would indeed be very helpful if `note.time_signature` were to access the current time signature (which might change at various points in the piece). I'm not sure if it's a good idea to actually store the data there, because then if you subsequently decide to change the time signature's location then you need to iterate over lots of notes and update them. Potentially better for the data to be stored in a form more like:

score.time_signatures = [
    TimeSignature("3/4", onset=0.0, offset=12.0),
    TimeSignature("4/4", onset=12.0, offset=24.0),
    TimeSignature("3/4", onset=24.0, offset=30.0),
]

and then the note attributes are computed dynamically from that? This is something that we could do pretty efficiently with my new `cached_event_property` code (see #68).

rbdannenberg:
I wouldn't suggest time or key signature as a per-note property, but if every note has a (likely) different property, it seems natural to store the information in the note object. Time signature and key signature are already implemented in basics.py and are inserted into the content of measures during musicxml and MIDI file input. You could argue these should be stored separately but it's a little tricky in that staves could have different key signatures or even time signatures. On the other hand, we *do* have a special Score property for tempo analogous to MIDI file tempo track (even though one could imagine per-staff independent tempo control.)

MarkGotham:
Thanks both. IMO
- at least some analysis objects (probably Meter, certainly Form) are going to have to be defined apart from notes, by reference to time positions alone. In case of interest, I wrote about this question in the [`When the Rome' paper (Gotham et al. 2023)](https://transactions.ismir.net/articles/10.5334/tismir.165/).
- my point of reference for how to store and navigate metrical _analysis_ is by a small modification to measure maps ... ah ... that's another, different case of [Gotham et al. 2023 ;)](https://dl.acm.org/doi/10.1145/3625135.3625136)
- Time signatures (nb: now we're talking about notation, not analysis) can happily live in measure objects. For reference, Music 21's solution for the functionality Peter mentions is 'get element by context' (measure, time signature, even some chord objects).
- this discussion has gotten a bit off track wrt Meter here ;)

pmcharrison:
Indeed, voice-specific time/key signatures complicate things. It makes sense, I suppose, to continue storing such notational objects within the `Score` streams, and then to use an equivalent of music21's `getContextByClass` to access the relevant time/key signature for a given note (thanks @MarkGotham for pointing out that method. Storing such objects in these stream naturally mirrors their location in the imported score data.

Presumably we don't want to store post hoc annotations in this form, though? Sometimes we will want to write something like this?

```py
score.annotations["metrical_hierarchies"] = Annotation([
    MetricalHierarchy(..., onset=0.0, offset=12.0),
    MetricalHierarchy(..., onset=12.0, offset=24.0),
    MetricalHierarchy(..., onset=24.0, offset=30.0),
])
```

and then write a helper function for `Note` that looks up the relevant metrical hierarchy:

```py
@property
def metrical_hierarchy(self):
    metrical_hierarchies = self.score.annotations["metrical_hiearchies"].locate(time=self.onset)
```

MarkGotham:
Tests will fail. Push simply to show design before proceeding.

pmcharrison:
Looking good also!

MarkGotham:
Finally (last back of commits) taken this as far as an algorithm!

MarkGotham:
TODO: handling of zero values with Gaussian.

amads/time/meter/attractor_tempos.py
```diff
@@ -0,0 +1,200 @@
...
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+# ------------------------------------------------------------------------------
+
+
+class MetricalSalience:
+    """
+    Organise array representations of metrical structre and derived salience values.
```
pmcharrison, line 25: typo

amads/time/meter/attractor_tempos.py
```diff
@@ -0,0 +1,200 @@
...
+
+    Parameters
+    ----------
+    symbolic_pulse_length_array: A NumPy array representing the symbolic pulse lengths.
```
pmcharrison, line 29: I think better just to call it `pulse_lengths` and leave type information to the type hints/documentation.

amads/time/meter/attractor_tempos.py
```diff
@@ -0,0 +1,200 @@
...
+
+    """
+
+    def __init__(
+        self,
+        symbolic_pulse_length_array: Optional[np.array] = None,
+    ):
+        self.symbolic_pulse_length_array = symbolic_pulse_length_array
+        self.pulse_lengths = self.symbolic_pulse_length_array[:, 0]
+        self.quarter_bpm = None
```
pmcharrison, line 66: I think you should consider choosing between either (1) Compute all these derived attributes in the `__init__` function or (2) don't store them as attributes at all but instead just return them as method outputs. The current solution feels like a strange middleground?

amads/time/meter/attractor_tempos.py
```diff
@@ -0,0 +1,200 @@
...
+           0.44793176, 1.4136999 , 0.44793176])
+
+    """
+
+    def __init__(
+        self,
+        symbolic_pulse_length_array: Optional[np.array] = None,
+    ):
+        self.symbolic_pulse_length_array = symbolic_pulse_length_array
+        self.pulse_lengths = self.symbolic_pulse_length_array[:, 0]
```
pmcharrison, line 65: The type hint says `None` is a valid value for `symbolic_pulse_length_array`, but `symbolic_pulse_length_array[:, 0]` will fail in such a case

amads/time/meter/attractor_tempos.py
MarkGotham: Ok. Debatable, but changed.

Done.

amads/time/meter/attractor_tempos.py
MarkGotham: Ok, I've tweaked all names and even pruned the number of attributes to this effect.

Done.

amads/time/meter/attractor_tempos.py
MarkGotham: Ok yes, fair enough. All on the init then (option 1). Probably not much of a use case for any of the intervening steps alone.

Done.

amads/time/meter/attractor_tempos.py
MarkGotham: True! Fixed.

Done.

amads/time/meter/attractor_tempos.py
pmcharrison: I meant that you had misspelled "structre"!

pmcharrison (approved):

amads/algorithms/__init__.py
```diff
@@ -19,7 +19,7 @@
 from ..polyphony.skyline import skyline
```
pmcharrison, line 19: As noted by email, maybe you could take this opportunity to remove `music21` from the list of dependencies in pyproject.toml?

amads/time/meter/attractor_tempos.py
MarkGotham: Oh in that case, not debatable!

:)

I wonder why the (usually overzealous) spell checker missed that.

amads/time/meter/attractor_tempos.py
MarkGotham: Great!

amads/algorithms/__init__.py
MarkGotham: Great, will do in a separate PR in case of unforeseen chaos ;)

PR #51, opened 2025-01-26
Implementations for melcontour and combcontour

diwangislucky:
Implementations for melcontour and combcontour from miditoolbox along with comments within the code. There's not much rationale to this implementation since it is relatively straightforward.


diwangislucky:
Not certain how I should resolve some of the errors in the tests when they aren't occurring inside the code I've contributed, or are we expected to get all of these tests working locally regardless?

amads/melody/contour/combcontour.py
```diff
@@ -0,0 +1,101 @@
...
+    Define the corresponding index of a note in a monophonic Score
+    as the order statistic of a note based off of note onset.
+    Define a combination contour matrix of a monophonic score as a boolean
+    square matrix whose row and column indices are the corresponding indices
+    of notes, and for all i,j-th element in the matrix such that (i > j and
+    note[j].keynum > note[i].keynum)
+    Computes a combination contour matrix given a monophonic Score.
+
+Dependencies:
+    - amads
```
pmcharrison, line 16: What's the motivation for listing dependencies here? I think it's a bit redundant and will. easily get out of date.

amads/melody/contour/combcontour.py
```diff
@@ -0,0 +1,101 @@
+"""
+Comparison melodic "contour" representation of a monophonic Score
+
+Date: [2025-01-26]
+
+Description:
+    Define the corresponding index of a note in a monophonic Score
```
pmcharrison, line 7: I'm not keen on putting key algorithm documentation in the module docstring, can't we just consolidate it all into the function docstring?

amads/melody/contour/melcontour.py
```diff
@@ -0,0 +1,164 @@
...
+        (you can perceive as (sampling_tick, pitch))
+
+    Notes
+    -----
+    Implementation based on the original MATLAB code from:
+    https://github.com/miditoolbox/1.1/blob/master/miditoolbox/melcontour.m
+
+    Examples
+    --------
+    [Unfortunately, still having problem resolving score construction]
```
pmcharrison, line 68: @Dav8Circle maybe worth having a look at this, and evaluating the overlap with your InterpolationContour implementation?

amads/melody/contour/combcontour.py
diwangislucky: that's fair. I will remove that.

amads/melody/contour/combcontour.py
diwangislucky: Okay. That sounds good! I will probably update the docstring to just be "A function to calculate the combination contour matrix"

amads/melody/contour/melcontour.py
diwangislucky: This is awkward. When I first worked on these functions InterpolationContour wasn't being worked on and I had the greenlight to work on this from Roger... If there's enough similarity, we could probably integrate our work or meet in the middle or something? In fact, I will also evaluate the overlap with InterpolationContour just to make sure.

diwangislucky: Just checked. Melcontour is a subcase of Salami. Ah well, duplicate work it is. I will leave it here just because I don't necessarily agree with salami.py using a custom class as an output, especially since core has a concurrent and sequential abstract representation class already and the custom class "Slice" does not derive from those things.

amads/melody/contour/combcontour.py
pmcharrison: OK, thanks!

amads/melody/contour/combcontour.py
pmcharrison: Great, thanks

amads/melody/contour/melcontour.py
pmcharrison: Sorry, I mixed up my terms; I meant to say that `melcontour` is basically the same as `StepContour` (see in particular the `_expand_to_vector` implementation in #27.

I hadn't thought of the overlap with salami.py. I don't think it's quite a subset of that, but I suppose that you could implement your algorithm using the `sliding_window` function from `slice/window.py`. It sounds a bit overcomplicated, though.

Dav8Circle: From the FANTASTIC docs (https://www.doc.gold.ac.uk/isms/m4s/FANTASTIC_docs.pdf), it seems to me that Müllensiefen would agree that `StepContour` and `melcontour` are essentially the same thing - indeed he cites the MIDI toolbox when defining Step Contour. 

`StepContour` is used to produce features `global_variation`, `global_direction` and `local_variation`, where `melcontour` returns the representation itself. In this way, I'm not sure whether the duplication is a problem? If a user wanted to obtain the aforementioned features from FANTASTIC, they ought to use `StepContour`. If they wanted to graph the contour representation, `melcontour` is the go-to (FANTASTIC makes no attempt to give the user direct access to the representation itself). I guess the question is whether it makes sense to make the user figure out which function/class is relevant to their needs, or consolidate it for simplicity?

 @pmcharrison not sure what is best for the user here!

pmcharrison: Thanks for this, @Dav8Circle! I think we definitely want to consolidate. I know that FANTASTIC doesn't give the user access to the representation, but that's no reason why we can't.

diwangislucky: Yes. I am down to help consolidate the codebase or do other gruntwork. Is there anything I need to check or anyone I need to talk to?

diwangislucky: > Thanks for this, @Dav8Circle! I think we definitely want to consolidate. I know that FANTASTIC doesn't give the user access to the representation, but that's no reason why we can't.

Also for this. I think, if we consolidate representations, we should probably inherit our classes from classes like sequence and concurrence in core.py to keep our internal representations consistent. If there are a ballooning amount of internal representations with no realistic consistency, we will encounter more increasing issues with trying to get multiple stages of score processing to work the more algorithms we implement.

pmcharrison: Thank you, @diwangislucky! 

I think the main issue that we have, really, is agreeing on I/O questions. There is already a good proposal that the input to @Dav8Circle's code should really be a `Score` object. @diwangislucky I'm not really sure what output class you are thinking of changing/updating, though? do you mean `StepContour` itself?

diwangislucky: So right now, we have roughly 3 modes of implementing things:
(1) salami.py uses the slice class in slice.py as an output, treating the output as something that we can't necessarily use in other places because they're specific to slice directory itself. While there's no particular reason why this is invalid per se, we have classes defined in core.py that achieve similar functionality (namely concurrence and sequence).
(2) treating the computation and its relevant initialization information as a first-class object. An example is stepcontour, in which we define a class in and of itself that holds both the computation and relevant state information to retrieve said computation. 
(3) Raw function from Score + needed control parameters to a raw interpretable output made of basic python data structures. Examples include pcdist1, segment_gestalt, etc.

While I understand that all these approaches are valid (save for a nitpick in the first one, in which case I'm inclined to believe that the output class should derive from core.py), it is an extra layer of complexity if we don't restrict what people are allowed to do for I/O.

diwangislucky: That being said, since large portions of this code-base has already been consolidated to, well, different avenues of implementation, and there are 10+ people working full speed on their own branches. I cannot really offer a good solution that requires minimal effort, so leave as is might be alright? I don't know.

pmcharrison: Hi @diwangislucky, thanks for these very useful points!

That's a very good point about the definition of the `Slice` class. I see no reason why `Slice` should not be modified to inherit from `Sequence`, and I should go ahead and do that.

From a UI perspective, I don't think (2) and (3) are that incompatible. They both involve invoking a function (which in (3) is an ordinary function, and in (2) is a constructor function), and they return an output, which may be basic Python objects (3) or a class containing multiple attributes (2). The most unpleasant aspect is that (2) uses CamelCase whereas (3) uses snake case, but I don't think that's a big deal really. I don't think the user experiences much additional complexity from this.

amads/melody/contour/combcontour.py
```diff
@@ -0,0 +1,95 @@
...
+        31(2), 225-267.
+"""
+
+import numpy as np
+
+from ...core.basics import Note, Score
+from ...pitch.ismonophonic import ismonophonic
+
+
+def combcontour(score: Score):
```
pmcharrison, line 27: We agreed in the style guide that we should try and go for consistent naming across the package, using snakecase and full words where possible. I appreciate that here you are trying to be consistent with MIDIToolbox but I think it's not so obvious to the user.

amads/melody/contour/melcontour.py
diwangislucky: Not necessarily I think? (2) allows for natural usage of lazy computation (since the code carrying out the computation is also part of the first-class object), and we can retrieve multiple computations from the same object (keep in mind it's computation, which allows for either up-front computation on initialization or lazy computation upon retrieval, maybe with additional state information, or some Frankenstein with methods in between). (3) forces a one-to-one correspondence between an input and output, promotes only up-front computation (there is no retrieval and initialization), and functions themselves aren't considered first-class objects in python I think? 

For the user, given that we're probably facing people who have less of a programming background, I think (3) will be the easiest and most intuitive paradigm, simply because it's simple to keep track of functionality in the sense that, "I'm only applying this thing once, up-front, and without the hassle of having to select between x different preprocessing/postprocessing methods. That being said, I may very well be wrong here since I honestly don't know what sort of user we're aiming the library at.

pmcharrison: But I think the problem is that (3) is fundamentally very limiting -- all the things you just described like up-front or lazy computation are very helpful when the operation is costly. I don't think (2) is that unusual a paradigm to use. See for example scikit-learn, where basically every algorithm is implemented this way.

pmcharrison: I don't think we can/should be targeting users for whom that is too complicated. Those users won't be able to make a plot with matplotlib or run a regression with statsmodels.

pmcharrison: Sorry about missing your issue, please feel free to continue the discussion on #56 as suggested.

Dav8Circle: I agree with Peter, and while I hear concerns about objects returning multiple computations, we can just write good documentation? Perhaps I am biased, but I think the StepContour docstring makes it quite clear how it is used! Plus, I don't think the beginner programmer you describe is really going to be concerned by whether we use lazy or up-front approaches - I certainly wouldn't have been when I started!

amads/melody/contour/combcontour.py
diwangislucky: Understood. I'll be consistent with this package.

PR #53, opened 2025-01-27
Implementing statistical expectation modelling

pmcharrison:


amads/expectation/model.py
```diff
@@ -0,0 +1,246 @@
...
+        entropy = 0
+        for _, p in distribution.distribution.items():
+            entropy -= p * math.log2(p)
+        return entropy
+    
+    def _calculate_max_entropy(self, support_size: int) -> float:
+        """Calculate maximum entropy Emax(X) for given support size."""
+        return math.log2(support_size)
+    
+    def _calculate_normalized_entropy(self, distribution: ProbabilityDistribution) -> float:
```
pmcharrison, line 161: This seems to duplicate methods in the ProbabilityDistribution class

amads/expectation/model.py
```diff
@@ -0,0 +1,246 @@
...
+        total_count = sum(self.ngrams[context].values()) + (self.smoothing_factor * V)
+        
+        # Calculate smoothed probability for all possible tokens
+        for token in self.vocabulary:
+            count = self.ngrams[context].get(token, 0)
+            probabilities[token] = (count + self.smoothing_factor) / total_count
+        
+        return ProbabilityDistribution(probabilities)
+
+class IDyOMModel(ExpectationModel):
```
pmcharrison, line 103: This is a very interesting and neat implementation; it doesn't capture all the details of IDyOM (especially viewpoints) but it gets many important aspects!

We could consider generalising this class to an `EnsembleModel` class which works by taking an arbitrary number of base models and combining their predictions in a specified manner. Real IDyOM has an interesting hiearchical structure, where IDyOM comprises the combination of STM and LTM, and each of STM and LTM comprise a combination of STM or LTM viewpoint models.

examples/plot_melodic_interval_expectation.py
```diff
@@ -0,0 +1,109 @@
...
+from amads.music import example
+from amads.expectation.tokenizer import MelodyIntervalTokenizer
+from amads.expectation.dataset import ScoreDataset
+from amads.expectation.model import MarkovModel, IDyOMModel
+from amads.expectation.metrics import NegativeLogLikelihood, Entropy
+import pickle
+
+# Get the path to the package's music directory-- to be fixed when sample datasets are properly included in the package
+package_dir = os.path.dirname(os.path.abspath(__file__))
+data_path = os.path.join(package_dir, '..', 'amads', 'music', 'Marion_2024_Bach_Chorales.pkl')
```
pmcharrison, line 19: This pkl format will be brittle, breaking if we change the Score class and associated classes (unless I'm missing something). Better instead to include a zip of MIDI files?

examples/plot_melodic_interval_expectation.py
```diff
@@ -0,0 +1,109 @@
...
+score_data = pickle.load(open(data_path, 'rb'))
+
+tokenizer = MelodyIntervalTokenizer()
+dataset = ScoreDataset(score_data, tokenizer)
+
+training_sequences = dataset[0:-10].copy()
```
pmcharrison, line 25: Why is `copy()` necessary?

examples/plot_melodic_interval_expectation.py
```diff
@@ -0,0 +1,109 @@
...
+
+# Create and train both models: A simple Markov model and a more complex IDyOM model
+idyom = IDyOMModel(max_order=5, smoothing_factor=0.01)
+markov = MarkovModel(order=5, smoothing_factor=0.01)
+
+idyom.train(training_sequences)
+markov.train(training_sequences)
+
+# Get predictions for test sequence
+idyom_predictions = idyom.predict_sequence(test_sequence)
```
pmcharrison, line 36: I wonder if it would be better to return a `Prediction` class rather than `ProbabilityDistribution` class? Then `Prediction` can contain attributes like `actual` (i.e. the actual token) as well as `nll` and `entropy`.

examples/plot_melodic_interval_expectation.py
```diff
@@ -0,0 +1,109 @@
...
+# as it progresses through a sequence? There are some ideas that this is
+# related to the reward people get from listening to music.
+# To implement this, we need to think about predict_sequence; maybe it's not
+# just predicting but also updating the model.
+#
+# from amads.expectation.metrics import cross_entropy
+
+# score = cross_entropy(predictions, tokens)
+
+# print(score)
```
pmcharrison, line 109: I think it would be wise to add some unit tests..... Cursor/ChatGPT can be very helpful for writing sensible tests

pmcharrison (commented):
Really good progress, well done! I have left various comments, I would be happy to discuss more when you like. The most subtle thing might be how to implement this EnsembleModel I mentioned, that could be something we work on together if you like.

amads/expectation/dataset.py
```diff
@@ -0,0 +1,123 @@
...
+    """Dataset implementation for Score input data that can handle both Score objects 
+    and MIDI file paths."""
+    
+    def __init__(self, 
+                 data: Union[List[Score], List[str], List[Path]], 
+                 tokenizer: Tokenizer):
+        """
+        Initialize the ScoreDataset.
+        
+        Args:
```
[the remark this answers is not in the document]
isebenius: Changed to numpydoc formatting! Will keep this in mind going forward

amads/expectation/dataset.py
```diff
@@ -0,0 +1,123 @@
...
+            data: List of Score objects or paths to MIDI files
+            tokenizer: Tokenizer instance for Score objects
+        """
+        self.source_paths: Dict[int, Optional[Path]] = {}
+        self.scores: List[Score] = []
```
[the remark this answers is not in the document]
isebenius: Fixed in https://github.com/music-computing/amads/commit/2aada0d50ba8cd2d7d470cbcec0e9940052c358f

amads/expectation/dataset.py
```diff
@@ -0,0 +1,123 @@
...
+        
+        # Process each input item
+        for idx, item in enumerate(data):
+            if isinstance(item, Score):
+                self.scores.append(item)
+                self.source_paths[idx] = None
```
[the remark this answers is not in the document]
isebenius: Agreed - this isn't really necessary here. Removed this extra metadata and source path information for the moment (https://github.com/music-computing/amads/commit/2aada0d50ba8cd2d7d470cbcec0e9940052c358f).

amads/expectation/dataset.py
```diff
@@ -0,0 +1,123 @@
...
+            else:
+                raise ValueError(f"Unsupported input type: {type(item)}")
+        
+        # Initialize the parent class with the processed scores
+        super().__init__(self.scores, tokenizer)
+    
+    def _load_midi(self, path: Path) -> Score:
+        """Load a MIDI file and convert it to a Score object."""
+        if not path.exists():
+            raise FileNotFoundError(f"MIDI file not found: {path}")
```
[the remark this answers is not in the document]
isebenius: Agreed, changed! https://github.com/music-computing/amads/commit/2aada0d50ba8cd2d7d470cbcec0e9940052c358f

examples/plot_melodic_interval_expectation.py
isebenius: Yeah definitely agreed -- I only put this here temporarily so I could run tests on the scores quickly and not have to call the `partitura_midi_import` function repeatedly since it was being very slow.

examples/plot_melodic_interval_expectation.py
isebenius: Yes, this makes sense! In 19d2ef2 I have created two new classes, `Prediction` and `SequencePrediction` which are backwards compatible with the current code (i.e. indexing them directly returns the predicted distribution) but also have other attributes including `observation,` `nll` and `entropy` -- hopefully this solutions works.

amads/expectation/model.py
isebenius: you are right -- removed these internal methods to only use the ones from the ProbabilityDistribution class.

amads/expectation/model.py
```diff
@@ -0,0 +1,246 @@
...
+            target = seq[i + self.order]
+            pred = self.predict_token(context, target)
+            predictions.append(pred)
+            
+        return predictions
+
+    def predict_token(self, context: tuple, current_token: Token) -> ProbabilityDistribution:
+        """Predict probability distribution for next token given context."""
+        # Add current_token to vocabulary if it's new
+        self.vocabulary.add(current_token)
```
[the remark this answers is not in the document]
isebenius: This makes sense -- i've changed it in d9b7a6e to be initialized at the outset of training as in the IDyOM model.

amads/expectation/model.py
```diff
@@ -0,0 +1,246 @@
...
+        """Generate predictions for each token in the sequence given previous tokens."""
+        predictions = []
+        seq = tuple(sequence)
+        
+        # For each position where we can make a prediction (starting after first note)
+        for i in range(len(seq) - 1):
+            # Train STM models on the sequence seen so far
+            current_sequence = seq[:i + 1]
+            for model in self.stm_models:
+                model.update([current_sequence])
```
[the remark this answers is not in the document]
isebenius: Yes this is a mistake here! I have addressed in 30209ed, which implements a distinct `update_online` function for just a single token, distinct from `update` which updates on a corpus.

amads/expectation/model.py
```diff
@@ -0,0 +1,246 @@
...
+            model_idx = order - 1
+            order_context = context[-order:]
+            
+            ltm_pred = self.ltm_models[model_idx].predict_token(order_context, target)
+            stm_pred = self.stm_models[model_idx].predict_token(order_context, target)
+            
+            ltm_predictions.append(ltm_pred)
+            stm_predictions.append(stm_pred)
+            
+        # Combine predictions using entropy weighting
```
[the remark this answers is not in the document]
isebenius: Thanks for these helpful comments about the IDyOM model! In ee9e64a I've updated the code to implement a few versions of the PPM algorithm which the user can choose, as well as the entropy method which I took from the recent "IDyOMpy" [paper](https://www.sciencedirect.com/science/article/pii/S0165027024002929). I still haven't implemented adding multiple viewpoints yet

amads/expectation/model.py
isebenius: Great idea! In 0eabe51 I've drafted a version of the `EnsembleModel` as well as child class `MarkovEnsemble` which is used as units within a newer implementation of then `IDyOMModel` class -- still doesn't capture viewpoints, but I think that should be simpler to implement, building off smaller ensemble model classes.

amads/expectation/tokenizers/midi_tokenizer.py
```diff
@@ -0,0 +1,272 @@
...
+                    if len(parts) == 3:
+                        num_beats = int(parts[0])
+                        num_samples = int(parts[1])
+                        resolution = int(parts[2])
+                        shift = num_beats + (num_samples / resolution)
+                        time_delta = shift * (60.0 / current_tempo)
+                        current_time += time_delta
+                elif token_str.startswith('Tempo_'):
+                    current_tempo = float(token_str.split('_')[1])
+            except Exception as e:
```
pmcharrison, line 264: Best not to catch and hide very general exceptions!

PR #57, opened 2025-01-27
Implement Score.get_notes()

pmcharrison:
Hi @rbdannenberg, am following up on our discussion here: https://github.com/music-computing/amads/pull/40#discussion_r1909222102

What do you think of this implementation? I thought also about just calling it `Score.notes` but I think if we are returning copies it's best to use `get_` so that it's clear that something is happening.

When writing the tests I saw some unexpected behaviour regarding tied notes -- currently the notes aren't being merged as I expect when the score is flattened. Would you mind having a look and confirming that this is unexpected behaviour?

### DIFF
--- a/amads/core/basics.py
+++ b/amads/core/basics.py
@@ -1007,6 +1007,10 @@ class Score(Concurrence):
         # Flatten to get selected notes in order of onset time
         return mtn.flatten(collapse=True)
 
+    def get_notes(self):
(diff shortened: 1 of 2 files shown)


pmcharrison:
Hi @rbdannenberg just checking if you saw this? We can resolve the question about ties in a separate issue.

pmcharrison:
Closed in favour of #68

PR #58, opened 2025-01-27
Implement Score.from_melody

pmcharrison:
An easy way to create a Score from a list of pitches and (optionally) deltas and durations.

Happy to revise the API if you have suggested improvements @rbdannenberg @Dav8Circle .

### DIFF
--- a/amads/core/basics.py
+++ b/amads/core/basics.py
@@ -35,7 +35,7 @@ Score (one per musical work or movement)
 import functools
 import weakref
 from math import floor
-from typing import List, Union
+from typing import List, Optional, Union
 
 from .time_map import TimeMap
 
@@ -829,19 +829,30 @@ class Score(Concurrence):
         self.time_map = time_map if time_map else TimeMap()
 
     @classmethod
-    def from_melody(cls, pitches: List[Union[int, Pitch]], deltas=None, durations=None):
-        """Create a Score from lists of pitches, deltas, and durations.
+    def from_melody(
+        cls,
+        pitches: List[Union[int, Pitch]],
+        durations: Optional[Union[float, List[float]]] = 1.0,
+        iois: Optional[Union[float, List[float]]] = None,
+        deltas: Optional[List[float]] = None,
+    ):
+        """Create a Score from a list of pitches and optional timing information.
+        Creates a monophonic melody where notes cannot overlap in time.
 
         Parameters
         ----------
-        pitches :
-            A list of pitches, specified either as MIDI note numbers or as Pitch objects.
-        deltas : list of float, optional
-            Start times in quarters relative to the score's start. If not provided,
-            defaults to [0, 1, 2, ...].
-        durations : list of float, optional
-            Durations in quarters for each note. If not provided, defaults to
-            1.0 for each note.
+        pitches : list of int or list of Pitch
+            MIDI note numbers or Pitch objects for each note
+        durations : float or list of float, optional
+            Durations in quarters for each note. If a scalar value, it will be repeated
+            for all notes. Defaults to 1.0 (quarter notes).
+        iois : float or list of float or None, optional
+            Inter-onset intervals in quarters between successive notes. If a scalar value,
+            it will be repeated for all notes. If not provided and deltas is None,
+            defaults to using the durations (notes placed sequentially without overlap).
+        deltas : list of float or None, optional
+            Start times in quarters relative to the melody's start. Cannot be used together
+            with iois. If both are None, defaults to using durations as IOIs.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


amads/core/basics.py
```diff
@@ -827,6 +827,83 @@ def __init__(self, delta=0, duration=0, content=None, time_map=None):
         super().__init__(delta, duration, content)
         self.time_map = time_map if time_map else TimeMap()
 
+    @classmethod
+    def from_melody(cls, pitches, deltas=None, durations=None):
+        """Create a Score from lists of pitches, deltas, and durations.
+
+        Parameters
+        ----------
+        pitches : list of int
```
rbdannenberg, line 836: It would make sense to allow a list of Pitch as well as list of int. Otherwise, you'll get default accidentals. There might also be a function somewhere that converts a list of ints to a list of Pitch given a key signature, e.g. 68 would become G# given keys of A or E or B but Bb with keys of F, Bb, Eb, etc.

amads/core/basics.py
```diff
@@ -827,6 +827,83 @@ def __init__(self, delta=0, duration=0, content=None, time_map=None):
...
+
+        Returns
+        -------
+        Score
+            A new Score object containing the melody in a single part
+
+        Raises
+        ------
+        ValueError
+            If input lists have different lengths or are empty
```
rbdannenberg, line 853: What if durations and deltas are in conflict?

pmcharrison: Mmm, good point. I agree, how about I replace deltas with iois? I think iois would probably be easier to use in most cases.

rbdannenberg: I agree, and maybe raise an error if deltas and iois are both provided? You could allow them to both be present, but with floating point rounding error, it's not clear what it means for these to be consistent, e.g. what if they differ by 1 microsecond?

pmcharrison: I agree, we can still support deltas as a possible argument, and I agree that it makes sense to throw an error if both are provided.

amads/core/basics.py
pmcharrison: Yes, good idea!

pmcharrison: Continuing in that direction, we could also support a list of strings e.g. C4, G4. Perhaps we also have a key_signature argument.

rbdannenberg: In the case of strings (which I have no problem with), we should use a string-to-Pitch conversion provided in the Pitch class (there may well be one already). I thought about key_signature argument, but I think it's a better design to implement list of ints with key_signature -> list of Pitch somewhere in the Pitch class where it might be more generally useful. Then you could call, e.g., Score.from_pitch(Pitch.from_int(range(60,67), key="F#maj"))

pmcharrison: > we should use a string-to-Pitch conversion provided in the Pitch class (there may well be one already)

agreed!

> I think it's a better design to implement list of ints with key_signature -> list of Pitch somewhere in the Pitch class where it might be more generally useful

Maybe we could still support this:

```py
Score.from_melody(pitches=range(60, 67), key="F#maj")
```

but under the hood we call this `Pitch.from_int` method? Maybe a little more user friendly than 

```py
Score.from_melody(pitches=Pitch.from_int(range(60, 67)), key="F#maj")
```

I suppose `Pitch.__init__` already takes an `int` as the `keynum` argument so it's not so obvious what's different about `Pitch.from_int`. As an alternative, we could consider adding an optional `key` argument to `Pitch.__init__`; the user could then choose to provide either of `alt` or `key` when creating a `Pitch` object. What do you think? I suppose we could consider other extensions to `Pitch.__init__` too; for example, it would be nice if people could create a `Pitch` from scientific notation, e.g.

```py
Pitch("C4")
Pitch(60)
Pitch(60, key="Cmaj")
Pitch(60, key="Bmaj")
Pitch(60, alt=1)
```

rbdannenberg: I agree with everything here. Pitch.__init__ should also take a list of ints and return a list if other functions are going to accept lists of Pitches (oh, so maybe you can't do that with __init__), and we need to define a notation for keys.

pmcharrison: Yep, perhaps a `Key` class but with accepted shorthand like `Cmaj` and `Cmin`? where the key has `tonic` and `mode` attributes, and `mode` is a string that can be used to index a reference dictionary of mode definitions?

amads/core/basics.py
pmcharrison: I've now implemented this, including various sensible defaults. See what you think?

amads/core/basics.py
pmcharrison: Have implemented the core proposed change here and have documented proposed extensions here: #61

amads/core/basics.py
rbdannenberg: Looks good, but why not allow for empty note lists creating an empty score?

pmcharrison: You're absolutely right, I've now implemented support for this. Let me know if you would like me to change anything else, otherwise I'll go ahead and merge.

rbdannenberg: go ahead and merge -- empty scores was the only question/comment I had.

pmcharrison: Thank you!

PR #60, opened 2025-01-30
Implement Melsim Wrapper

Dav8Circle:
Collaborative coding work with @pmcharrison to wrap the _melsim_ package in R: https://github.com/sebsilas/melsim
Implements the wrapper itself in `/melody/similarity/melsim.py` and a demo in `demos/melsim.py`.
Top level function `get_similarity` takes the list of melodies (`Score` objects) to be compared, and a string corresponding to one of the similarity measures listed in the docstring (case-sensitive). Returns the given similarity score for each pairwise melody comparison

### DIFF
--- a/amads/melody/similarity/melsim.py
+++ b/amads/melody/similarity/melsim.py
@@ -76,6 +76,9 @@ from amads.utils import check_python_package
 base_packages = ["base", "utils"]
 cran_packages = ["tibble", "R6", "remotes"]
 github_packages = ["melsim"]
+github_repos = {
+    "melsim": "sebsilas/melsim",
+}
 
 R = SimpleNamespace()
 
@@ -100,30 +103,39 @@ def requires_melsim(func):
     return wrapper
 
 
-def check_packages_installed():
-    from rpy2.robjects.packages import importr, isinstalled
+def check_packages_installed(install_missing: bool = False):
+    from rpy2.robjects.packages import isinstalled
 
-    utils = importr("utils")
-    remotes = importr("remotes")
-    for package in cran_packages:
+    for package in cran_packages + github_packages:
         if not isinstalled(package):
-            response = input(
-                f"Package '{package}' is not installed. Would you like to install it? (y/n): "
-            )
-            if response.lower() != "y":
-                raise ImportError(f"Package '{package}' is required but not installed")
+            if install_missing:
+                install_r_package(package)
+            else:
+                raise ImportError(
+                    f"Package '{package}' is required but not installed. "
+                    "You can run install it by running the following command: "
+                    "from amads.melody.similarity.melsim import install_dependencies; install_dependencies()"
+                )
 
-            utils.install_packages(package)
 
-    for package in github_packages:
-        if not isinstalled(package):
-            response = input(
-                f"Package '{package}' is not installed. Would you like to install it? (y/n): "
-            )
-            if response.lower() != "y":
-                raise ImportError(f"Package '{package}' is required but not installed")
+def install_r_package(package: str):
+    from rpy2.robjects.packages import importr
 
-            remotes.install_github(package)
+    if package in cran_packages:
+        print(f"Installing CRAN package '{package}'...")
+        utils = importr("utils")
+        utils.install_packages(package)
+    elif package in github_packages:
+        print(f"Installing GitHub package '{package}'...")
+        remotes = importr("remotes")
+        repo = github_repos[package]
+        remotes.install_github(repo, upgrade="always")
+    else:
+        raise ValueError(f"Unknown package type for '{package}'")
+
+
+def install_dependencies():
+    check_packages_installed(install_missing=True)
 
 
 def import_packages():
--- a/amads/melody/similarity/melsim.py
+++ b/amads/melody/similarity/melsim.py
@@ -125,11 +125,13 @@ def install_r_package(package: str):
         print(f"Installing CRAN package '{package}'...")
         utils = importr("utils")
         utils.install_packages(package)
+
     elif package in github_packages:
         print(f"Installing GitHub package '{package}'...")
-        remotes = importr("remotes")
         repo = github_repos[package]
-        remotes.install_github(repo, upgrade="always")
+        devtools = importr("devtools")
+        devtools.install_github(repo, upgrade="always")
+
     else:
         raise ValueError(f"Unknown package type for '{package}'")
 
--- a/amads/melody/similarity/melsim.py
+++ b/amads/melody/similarity/melsim.py
@@ -205,8 +205,8 @@ def get_similarity(
     >>> melody_2 = Score.from_melody(pitches=[60, 62, 64, 67], durations=1.0)
     >>> # Calculate similarity using Jaccard method
     >>> similarity = get_similarity(melody_1, melody_2, 'Jaccard', 'pitch')
-    >>> similarity == 0.6 # verified against melsim R package
-    True
+    >>> similarity # obtained from R package
+    0.6
     >>> # Verify that similarity between the same melody is 1
     >>> similarity = get_similarity(melody_1, melody_1, 'Jaccard', 'pitch')
     >>> similarity == 1
--- a/amads/melody/similarity/melsim.py
+++ b/amads/melody/similarity/melsim.py
@@ -1,7 +1,75 @@
+"""
+This is a Python wrapper for the R package 'melsim' (https://github.com/sebsilas/melsim).
+This wrapper seeks to allow the user to easily interface with the melsim package using
+the Score objects in AMADS.
+
+Melsim is a package for computing similarity between melodies. It is based on SIMILE, which was
+written by Daniel Müllensiefen and Klaus Frieler in 2003/2004. Melsim is used to compare two
+or more melodies pairwise across a range of similarity measures. Not all similarity measures
+are implemented in melsim, but the ones that are can be used in AMADS.
+
+All of the following similarity measures are implemented and functional in melsim:
+Please be aware that the names of the similarity measures are case-sensitive.
+
+Num:        Name:
+1           Jaccard
+2       Kulczynski2
+3            Russel
+4   simple matching
+5             Faith
+6          Tanimoto
+7              Dice
+8            Mozley
+9            Ochiai
+10          Simpson
+11           cosine
+12          angular
+13      correlation
+14        Tschuprow
+15           Cramer
+16            Gower
+17        Euclidean
+18        Manhattan
+19         supremum
+20         Canberra
+21            Chord
+22         Geodesic
+23             Bray
+24          Soergel
+25           Podani
+26        Whittaker
+27         eJaccard
+28            eDice
+29   Bhjattacharyya
+30       divergence
+31        Hellinger
+32    edit_sim_utf8
+33         edit_sim
+34      Levenshtein
+35          sim_NCD
+36            const
+37          sim_dtw
+
+The following similarity measures are not currently functional in melsim:
+1    count_distinct (set-based)
+2          tversky (set-based)
+3   braun_blanquet (set-based)
+4        minkowski (vector-based)
+5           ukkon (distribution-based)
+6      sum_common (distribution-based)
+7       distr_sim (distribution-based)
+8   stringdot_utf8 (sequence-based)
+9             pmi (special)
+10       sim_emd (special)
+"""
+
 from functools import cache
+from itertools import combinations
 from types import SimpleNamespace
 
-from amads.core.basics import Score
+import numpy as np
+
+from amads.core.basics import Note, Score
 from amads.pitch.ismonophonic import ismonophonic
 from amads.utils import check_python_package
 
@@ -63,15 +131,43 @@ def import_packages():
 
     all_packages = base_packages + cran_packages + github_packages
     for package in all_packages:
-        R[package] = importr(package)
+        setattr(R, package, importr(package))
 
 
 @requires_melsim
-def get_similarity(melody_1: Score, melody_2: Score, method: str):
-    pass_melody_to_r(melody_1, "melody_1")
-    pass_melody_to_r(melody_2, "melody_2")
+def get_similarity(melodies: list[Score], method: str) -> dict:
+    """
+    Calculate pairwise similarities between all melodies using the specified method.
+
+    Args:
+        melodies: List of Score objects containing monophonic melodies
+        method: Name of the similarity method to use
+
+    Returns:
+        Dictionary containing pairwise similarities with normal float values
+    """
+    n = len(melodies)
+    similarities = np.zeros((n, n))
+
+    # Load all melodies into R
+    for i, melody in enumerate(melodies):
+        pass_melody_to_r(melody, f"melody_{i + 1}")
+
+    # Load the similarity measure
     load_similarity_measure(method)
-    return _get_similarity("melody_1", "melody_2", method)
+    # Calculate pairwise similarities
+    for i, j in combinations(range(n), 2):
+        sim = _get_similarity(f"melody_{i + 1}", f"melody_{j + 1}", method)
+        similarities[i, j] = sim
+        similarities[j, i] = sim
+
+    # Convert to dictionary of melody pairs and similarity values
+    similarity_dict = {}
+    for i, j in combinations(range(n), 2):
+        key = (f"melody_{i + 1}", f"melody_{j + 1}")
+        similarity_dict[key] = float(similarities[i, j])
+
+    return similarity_dict
 
 
 loaded_melodies = {}
@@ -87,6 +183,7 @@ def pass_melody_to_r(melody: Score, name: str):
         A melsim Melody object
     """
     import rpy2.robjects as ro
+    from rpy2.robjects import FloatVector
 
     if name in loaded_melodies and loaded_melodies[name] is melody:
         return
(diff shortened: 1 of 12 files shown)


Dav8Circle:
@pmcharrison It seems our approach to give the user the choice whether to install `rpy2` etc. is not well liked by our test suite, nor codecov! We may have to rethink...

pmcharrison:
> @pmcharrison It seems our approach to give the user the choice whether to install `rpy2` etc. is not well liked by our test suite, nor codecov! We may have to rethink...

The way to do this is to put the optional packages in `pyproject.toml` as a dependency group called `melsim`, and then update .github/workflows/tests.yml so that when it pip installs the package it installs the melsim dependencies too. Cursor can help you with that, I reckon.

pmcharrison:
PS you will probably have to update tests.yml to install R also.

pmcharrison:
In fact, thinking out loud, I suppose tests.yml should be responsible for installing remotes and melsim.

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,290 @@
...
+            name=method,
+            full_name=method,
+            transformation="pitch",
+            parameters=ro.ListVector({}),
+            sim_measure=method,
+        ),
+    )
+
+
+def _get_similarity(melody_1: str, melody_2: str, method: str):
```
pmcharrison, line 274: Needs `@requires_melsim` decorator

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,290 @@
...
+        key = (f"melody_{i + 1}", f"melody_{j + 1}")
+        similarity_dict[key] = float(similarities[i, j])
+
+    return similarity_dict
+
+
+loaded_melodies = {}
+
+
+def pass_melody_to_r(melody: Score, name: str):
```
pmcharrison, line 225: Consider prefixing with underscore to make internal

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,290 @@
...
+Melsim is a package for computing similarity between melodies. It is based on SIMILE, which was
+written by Daniel Müllensiefen and Klaus Frieler in 2003/2004. Melsim is used to compare two
+or more melodies pairwise across a range of similarity measures. Not all similarity measures
+are implemented in melsim, but the ones that are can be used in AMADS.
+
+All of the following similarity measures are implemented and functional in melsim:
+Please be aware that the names of the similarity measures are case-sensitive.
+
+Num:        Name:
+1           Jaccard
```
pmcharrison, line 15: I see how you can choose between different list similarity metrics here, do we have any options to choose what melodic representation is used for those computations? I thought that was half of what SIMILE was all about (deriving representations from melodies).

pmcharrison (commented):
I have fixed the CI, made it much faster to run, and separated out the melsim tests into their own CI job to keep it from slowing down the Python tests.

pmcharrison:
@Dav8Circle FYI #58 (`Score.from_melody`) is now merged so you can make use of that now?

Dav8Circle:
> @Dav8Circle FYI #58 (`Score.from_melody`) is now merged so you can make use of that now?

Awesome, thanks!

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,241 @@
...
+def import_packages():
+    from rpy2.robjects.packages import importr
+
+    all_packages = base_packages + cran_packages + github_packages
+    for package in all_packages:
+        setattr(R, package, importr(package))
+
+
+@requires_melsim
+def get_similarity(melodies: list[Score], method: str) -> dict:
```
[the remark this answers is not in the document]
Dav8Circle: The idea here was that for large lists of melodies, where pairwise comparisons are desired for every combination, that's going to be a lot of unnecessary loadings of the same melody. If we instead support the matrix approach from the get-go, then each melody need only be created in R once and we can reuse the same instance of the object as needed. That was my thinking at least - not sure whether it's implemented in the most effective way to that goal

amads/melody/similarity/melsim.py
Dav8Circle: fixed, thanks

amads/melody/similarity/melsim.py
Dav8Circle: yeah fair enough, fixed!

amads/melody/similarity/melsim.py
Dav8Circle: As soon as I figure out exactly what melsim is doing, I'll try and support that - more forensic work required in the absence of docs!

pmcharrison: Yeah, feel free to pop Seb Silas a message too (sebsilas@googlemail.com). He would be eager to discuss I'm sure!

pmcharrison: Yes, you could contact Seb Silas too, I'm sure he'd be happy to help!

amads/melody/similarity/melsim.py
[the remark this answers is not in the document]
pmcharrison: I see, yes -- while the `loaded_melodies` functionality partly helps with that, it doesn't help entirely, because only two melodies are ever stored in memory. Fair enough!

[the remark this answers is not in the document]
pmcharrison: If I were doing it, I might suggest something like this:

- Expose `get_similarity` as the recommended interface for most users
- Remove `get_similarities` (matrix version)
- Expose `_pass_melody_to_r` but call it `r_load_melody`
- Expose `_get_similarity` but call it `r_get_similarity`
- In `get_similarity` docstring, explain that for performance-intensive applications where the same melody is being reused multiple times, users can get better performance by using `r_load_melody` and `r_get_similarity`

Dav8Circle:
@pmcharrison I've implemented what you suggested: `get_similarity` is the recommended interface for simply comparing two melodies, and the matrix version has been moved to `r_get_similarity`, which I provide example usage for in `demos`

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
+"""
+This is a Python wrapper for the R package 'melsim' (https://github.com/sebsilas/melsim).
+This wrapper seeks to allow the user to easily interface with the melsim package using
+the Score objects in AMADS.
```
pmcharrison, line 4: Can you credit Seb Silas too?

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
...
+"""
+
+from functools import cache
+from types import SimpleNamespace
+
+from amads.core.basics import Note, Score
+from amads.pitch.ismonophonic import ismonophonic
+from amads.utils import check_python_package
+
+base_packages = ["base", "utils"]
```
pmcharrison, line 73: can we rename as `r_base_packages` and `r_cran_packages` and `r_github_packages`?

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
...
+
+@cache
+def load_melsim():
+    check_python_package("pandas")
+    check_python_package("rpy2")
+
+    from rpy2.robjects import pandas2ri
+
+    pandas2ri.activate()
+    check_packages_installed()
```
pmcharrison, line 92: Can we rename as `check_r_packages_installed`?

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
...
+    import_packages()
```
pmcharrison, line 93: can we rename as `check_r_packages_installed`?

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
...
+github_repos = {
+    "melsim": "sebsilas/melsim",
+}
+
+R = SimpleNamespace()
+
+
+@cache
+def load_melsim():
+    check_python_package("pandas")
```
pmcharrison, line 86: Can we rename as `check_python_package_installed` (that will involve renaming in utils too)

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
...
+
+    from rpy2.robjects import pandas2ri
+
+    pandas2ri.activate()
+    check_packages_installed()
+    import_packages()
+
+
+def requires_melsim(func):
+    from functools import wraps
```
pmcharrison, line 97: can we make this a global import?

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
...
+
+    Examples
+    --------
+    >>> from amads.core.basics import Score
+    >>> # Create two simple melodies using from_melody
+    >>> melody1 = Score.from_melody(pitches=[60, 62, 64, 65], durations=1.0)
+    >>> melody2 = Score.from_melody(pitches=[60, 62, 64, 67], durations=1.0)
+    >>> # Calculate similarity using Jaccard method
+    >>> similarity = get_similarity(melody1, melody2, 'Jaccard')
+    >>> 0 <= similarity <= 1
```
pmcharrison, line 176: Make this a strict inequality

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
...
+    True
```
pmcharrison, line 177: Add another example showing that the similarity of melody1 with itself is 1?

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
...
+def import_packages():
+    from rpy2.robjects.packages import importr
+
+    all_packages = base_packages + cran_packages + github_packages
+    for package in all_packages:
+        setattr(R, package, importr(package))
+
+
+@requires_melsim
+def get_similarity(melody1: Score, melody2: Score, method: str) -> float:
```
pmcharrison, line 151: I think we should rename `melody1` to `melody_1`

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,267 @@
...
+    Returns
+    -------
+    float
+        Similarity value between the two melodies
+
+    Examples
+    --------
+    >>> from amads.core.basics import Score
+    >>> # Create two simple melodies using from_melody
+    >>> melody1 = Score.from_melody(pitches=[60, 62, 64, 65], durations=1.0)
```
pmcharrison, line 172: I think we should rename `melody1` to `melody_1`

pmcharrison (commented):
Looking very good, just small comments now!

amads/melody/similarity/melsim.py
Dav8Circle: Done!

amads/melody/similarity/melsim.py
Dav8Circle: Done

amads/melody/similarity/melsim.py
Dav8Circle: yep

amads/melody/similarity/melsim.py
Dav8Circle: fixed

amads/melody/similarity/melsim.py
Dav8Circle: done

amads/melody/similarity/melsim.py
Dav8Circle: done

amads/melody/similarity/melsim.py
Dav8Circle: done

amads/melody/similarity/melsim.py
Dav8Circle: added

amads/melody/similarity/melsim.py
Dav8Circle: done

amads/melody/similarity/melsim.py
Dav8Circle: done

amads/melody/similarity/melsim.py
```diff
@@ -3,10 +3,12 @@
 This wrapper seeks to allow the user to easily interface with the melsim package using
 the Score objects in AMADS.
 
-Melsim is a package for computing similarity between melodies. It is based on SIMILE, which was
-written by Daniel Müllensiefen and Klaus Frieler in 2003/2004. Melsim is used to compare two
-or more melodies pairwise across a range of similarity measures. Not all similarity measures
-are implemented in melsim, but the ones that are can be used in AMADS.
+Melsim is a package for computing similarity between melodies, and was written by
```
pmcharrison, line 6: Is it fair to say "was written by Sebastian"? Not sure I would read that from Klaus's email. Better to write "is being developed by Sebastian Silas and Klaus Frieler"?

amads/melody/similarity/melsim.py
pmcharrison: I think basically we need a bit more clarity on how the available algorithms operate, e.g. are they restricted currently to the pitch domain, do they transform pitch at all before performing the computations, etc. We could also indicate where we expect the package to develop in the future, i.e. what kind of similarity computations will become available eventually.

Dav8Circle:
Also, any ideas on how to fix the tests again? Seems like a small thing with codecov but not sure how to work that yet

amads/melody/similarity/melsim.py
```diff
@@ -0,0 +1,324 @@
...
+        If the number of melodies is not exactly two.
+
+    Examples
+    --------
+    >>> from amads.core.basics import Score
+    >>> # Create two simple melodies using from_melody
+    >>> melody_1 = Score.from_melody(pitches=[60, 62, 64, 65], durations=1.0)
+    >>> melody_2 = Score.from_melody(pitches=[60, 62, 64, 67], durations=1.0)
+    >>> # Calculate similarity using Jaccard method
+    >>> similarity = get_similarity(melody_1, melody_2, 'Jaccard')
```
pmcharrison, line 207: Presumably now all these get_similarity calls need to specify the representation?

pmcharrison:
> @pmcharrison I hope that looks a bit better - have worked out what the transformations are and which ones currently work! Can't help but feel as though the docstring has become a bit of a wall-of-text, and none of the measures or transformations are actually explained - from a user POV, I feel we probably ought to explain what these things are and how they work, but not sure how best to approach it. Ideally, we'd just cite a paper, but I don't believe one has been written yet?

Awesome, this really helps! Sure I think one would ideally document those things more explicitly but I think it's moderately obvious from the names. You're right that long-term we can then point to the melsim documentation.

pmcharrison:
> Also, any ideas on how to fix the tests again? Seems like a small thing with codecov but not sure how to work that yet

Sorry I broke that, I should have fixed it now

amads/melody/similarity/melsim.py
Dav8Circle: I've set `"pitch"` as a default in `load_similarity_measure` so didn't notice this, but yes!

Dav8Circle:
Just to summarise the testing problem:

the Jaccard sim (pitch) for melody_1 and melody_2 is 0.6, validated from original melsim package and local run.
when run through CI, this inexplicably returns 1.0

This is why the initial iteration of the test (`0 < similarity < 1`) failed. It is completely unbeknownst to me why the CI returns a different value for the test

pmcharrison:
Heheh, how mysterious...

My recommendation: ask Cursor to help you run those tests locally using Docker, with a view to reproducing the exact test environment that produces this weird error. You can then get it to provide a breakpoint so you can explore interactively what's going on.

Dav8Circle:
I've had a go with a couple of different approaches - using a completely fresh venv, and the Docker approach. Both yield the correct value of 0.6! Will continue trying to reproduce this error tomorrow morning!

pmcharrison (approved):

tests/test_melsim.py
```diff
@@ -0,0 +1,92 @@
...
+
+    assert callable(get_similarity)
+
+
+def test_example_usage():
+    mel_1 = Score.from_melody(pitches=[60, 62, 64, 65], durations=1.0)
+    mel_2 = Score.from_melody(pitches=[60, 62, 64, 67], durations=1.0)
+    similarity = get_similarity(mel_1, mel_2, "Jaccard", "pitch")
+    assert similarity == 0.6
+
```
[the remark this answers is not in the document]
Dav8Circle: Added!

rbdannenberg:
melsim.py has this code:
                   raise ImportError(
                        f"Package '{package}' is required but not installed. "
                        "You can run install it by running the following command: "
                        "from amads.melody.similarity.melsim import install_dependencies; install_dependencies()"
                    )
but if you run it from a test, pytest truncates the message, and it seemed to me that the test was missing Python package (tibble) that fortunately does not exist for Python, or it would be *really* confusing. Eventually I searched the codebase for "tibble", found this message and got pytest to work. It's also a huge and slow-to-install dependency (265MB in homebrew). Are we planning to rewrite the R dependency in Python?

pmcharrison:
Hi @rbdannenberg, sorry you had a bad experience here!

With pytest, if you see the error message truncated, you have to scroll up to see the full one. This is what I get if I reproduce your problem by uninstalling tibble:

image

It'll be more obvious once we sort those warning messages in between.

If you had seen the full error message:

```
ImportError: Package 'tibble' is required but not installed. You can run install it by running the following command: from amads.melody.similarity.melsim import install_dependencies; install_dependencies()
```

My goal was that, if you do see the proper error message, it should be straightforward to follow.

It's worth providing some context here, that I have provided some infrastructure for optional dependencies in amads, such that there are particular modules whose dependencies won't install by default, typically because they are heavy or place additional platform constraints. Melsim is one of those modules; it is requiring an R installation and the associated melsim R package. The idea is that these modules shouldn't obstruct ordinary package usage, and I think I've achieved that in most cases, but your problematic experience indicates to me that pytest should by default skip such modules when run locally.

> Also, before I even got this far, pytest will hang when tests/test_demos.py runs demos/melsim.py -- I had to control-Z and use kill to regain control and isolate the problem.

That doesn't sound good, it's not a problem that I've experienced on my end, with or without dependencies installed. I will see if I can reproduce it.

> It's also a huge and slow-to-install dependency (265MB in homebrew). Are we planning to rewrite the R dependency in Python?

Yes, it's a big dependency, I think it's important that it stays strictly optional and is not installed by default. I think it would be good to have this in Python ultimately, but I know that the original package authors (Silas and Frieler) are still working away at their R implementation, and I think it's best to keep the efforts consolidated for the time being.

PR #63, opened 2025-02-07
Implement melody analysis functions

yiwenzcmu:
implement melodic functions according to MIDI Toolbox manual. Reference: [https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=6e06906ca1ba0bf0ac8f2cb1a929f3be95eeadfa#page=50](url)


pmcharrison:
@yiwenzcmu I quickly looked through, these look great! I will initialise a few comment threads to discuss particular details, I'll tag @rbdannenberg and @MarkGotham too in case they would like to chip in.

amads/melody/compl_trans.py
```diff
@@ -0,0 +1,167 @@
...
+    
+    # Convert to originality score (inverse probability scaled to 0-10)
+    # Adjust scaling to get desired range
+    originality = 10 * (1 - (avg_probability / 0.4))
+    originality = max(0, min(10, originality))
+    
+    return originality
+
+
+if __name__ == "__main__":
```
pmcharrison, line 112: Put these kinds of scripts in demos instead?

amads/melody/gradus.py
```diff
@@ -0,0 +1,142 @@
+__author__ = "Yiwen Zhao"
+
+import numpy as np
+from ..core.basics import Score, Note, Part
+
+def calculate_gradus(score: Score) -> float:
```
pmcharrison, line 6: Shorten functions like `calculate_` to `get_` to save typing?

amads/melody/mobility.py
```diff
@@ -0,0 +1,112 @@
...
+    Parameters
+    ----------
+    score : Score
+        A Score object containing the melody to analyze. The score will be
+        flattened and collapsed into a single sequence of notes ordered by
+        onset time.
+    
+    Returns
+    -------
+    numpy.ndarray
```
pmcharrison, line 23: Earlier we agreed that we should avoid returning numpy types, and stay consistent with basic Python types (e.g. lists of floats). Are we still happy with that @rbdannenberg, @MarkGotham? I do see that it's a bit ugly to coerce to list right at the end, but I also think it's a bit ugly to be inconsistent within our functions. I suppose we could move everything to numpy, but that is a bit of a leap too.

amads/melody/narmour.py
```diff
@@ -0,0 +1,208 @@
...
+    part = Part()
+    notes = [
+        Note(pitch=60),  # C4
+        Note(pitch=64),  # E4
+        Note(pitch=67),  # G4
+        Note(pitch=72)   # C5
+    ]
+    for note in notes:
+        part.append(note)
+    score.append(part)
```
pmcharrison, line 170: A lot of these kinds of passages can now be simplified using the new `Score.get_notes` method.

amads/melody/mel_accent.py
```diff
@@ -0,0 +1,129 @@
...
+    ...     part.append(note)
+    >>> score.append(part)
+    >>> accent_values = calculate_melodic_accent(score)
+    >>> print(accent_values)
+    [0.0, 1.0, 0.0]  # Peak accent in the middle
+    """
+    # Flatten and collapse the score into a single sequence of notes
+    flattened_score = score.flatten(collapse=True)
+    
+    # Handle empty score
```
pmcharrison, line 47: I don't think we need to deal with this case explicitly, as it is already covered by l. 48-49.

amads/melody/mel_accent.py
```diff
@@ -0,0 +1,129 @@
...
+    >>> notes = [Note(pitch=60), Note(pitch=64), Note(pitch=62)]  # up-down pattern
+    >>> for note in notes:
+    ...     part.append(note)
+    >>> score.append(part)
+    >>> accent_values = calculate_melodic_accent(score)
+    >>> print(accent_values)
+    [0.0, 1.0, 0.0]  # Peak accent in the middle
+    """
+    # Flatten and collapse the score into a single sequence of notes
+    flattened_score = score.flatten(collapse=True)
```
pmcharrison, line 45: I think we should check if the score is monophonic and throw an error if not

amads/melody/mel_attraction.py
```diff
@@ -0,0 +1,197 @@
...
+    >>> part = Part()
+    >>> notes = [Note(pitch=60), Note(pitch=64), Note(pitch=67), Note(pitch=65)]
+    >>> for note in notes:
+    ...     part.append(note)
+    >>> score.append(part)
+    >>> attractions = calculate_melodic_attraction(score, key=60)
+    >>> print(attractions)  # Example values
+    [0.0, 0.45, 0.8, 0.6]
+    """
+    flattened_score = score.flatten(collapse=True)
```
pmcharrison, line 109: I think we should check if the score is monophonic and throw an error if not

amads/melody/mel_distance.py
```diff
@@ -0,0 +1,64 @@
...
+    -------
+    float
+        Value representing the distance between the two scores under the given representation and metric.
+
+    Examples
+    --------
+    >>> calculate_mel_distance(score1, score2, 'pcdist1', 'taxi')
+    0.16666666666666666  # Example output
+    """
+    if repr == 'pcdist1':
```
pmcharrison, line 36: I think we should check if the score is monophonic and throw an error if not

amads/melody/meteraccent.py
```diff
@@ -0,0 +1,193 @@
...
+    ...     Note(pitch=67, dur=1.0)   # Quarter note G4
+    ... ]
+    >>> for note in notes:
+    ...     part.append(note)
+    >>> score.append(part)
+    >>> accent_sync = calculate_meteraccent(score)
+    >>> print(accent_sync)
+    0.75
+    """
+    def calculate_durational_accents(notes: list) -> np.ndarray:
```
pmcharrison, line 53: Are you sure you want to define these functions locally? Why not put them in the general module namespace? You could prefix with underscore if you want them to be private.

amads/melody/mobility.py
```diff
@@ -0,0 +1,112 @@
...
+    ...     part.append(note)
+    >>> score.append(part)
+    >>> mobility_values = calculate_mobility(score)
+    >>> print(mobility_values)
+    array([nan, -0.5,  0.4, -0.4, nan])
+    """
+
+    # Flatten and collapse the score into a single sequence of notes
+    flattened_score = score.flatten(collapse=True)
+
```
pmcharrison, line 43: I think we should check if the score is monophonic and throw an error if not

amads/melody/narmour.py
```diff
@@ -0,0 +1,208 @@
...
+    >>> part = Part()
+    >>> notes = [Note(pitch=60), Note(pitch=64), Note(pitch=67)]
+    >>> for note in notes:
+    ...     part.append(note)
+    >>> score.append(part)
+    >>> rd_values = calculate_narmour(score, 'rd')
+    >>> print(rd_values)
+    array([nan, nan, 0.75])
+    """
+    def registral_direction(pitches: np.ndarray) -> np.ndarray:
```
pmcharrison, line 62: I think these should be defined at the top level of the module

amads/melody/narmour.py
```diff
@@ -0,0 +1,208 @@
...
+    # Calculate expectations based on selected principle
+    principle_functions = {
+        'rd': registral_direction,
+        'rr': registral_return,
+        'id': intervallic_difference,
+        'cl': closure,
+        'pr': proximity,
+        'co': consonance
+    }
+
```
pmcharrison, line 153: Need to check that it's monophonic

amads/melody/mobility.py
rbdannenberg: @pmcharrison We talked this morning and my opinion is that if the analysis returns a value for each note, the value should be added as a note attribute. To keep the immutability property, I think the score should be copied before changing it, although the voice in my head is saying that treating the score as a "blackboard" that gets updated with new information might be better. I guess for now, I'm sticking to my guns and saying make a copy until there's a demonstrated need for the "blackboard" approach.

pmcharrison: @dav and I discussed a similar matter recently. The difficulty I had is that it's considered a sin in Python to initialise object attributes outside `__init__`. What to do then? Some options:

1. Initialise all these possible analysis attributes in `Note.__init__`
2. Create a dot-dictionary called `data`, so we get e.g. `note.data.ioi`
3. Initialise the object attributes wherever we want
4. Don't store these attributes in notes after all, since it would yield an inconsistent API where some functions return scores and some functions return other kinds of scalars or lists

amads/melody/mel_accent.py
rbdannenberg: If monophonic means strictly non-overlapping, then almost any melody recorded on a MIDI keyboard is likely to fail the monophonic test due to slightly overlapping notes.  It might still be best to force users to explicitly convert to monophonic non-overlapping -- at least they won't be surprised when a clearly chordal performance sneaks in and gives confusing results.

pmcharrison: Hmm, that's a good point. I suppose many of these algorithms will work as intended with such a melody (i.e. one with slight overlaps)? In such cases, maybe my `is_monophonic` test would do more harm than good. Maybe it's the user's fault if they run it on something stupid...

amads/melody/mobility.py
rbdannenberg: I think the "original sin" :-) was the Python design, but this is one case where the fact that "objects" in Python are really dictionaries makes it an ideal way represent an extensible structure. You could pretend objects are not extensible, and my non-Python music representation libraries add a properties field to events where properties is a list of attribute/value pairs, just like the note.data.ioi idea, but I can't see the advantage of that since notes are already essentially dictionaries.

pmcharrison: These arguments make sense and we should make something like that work. I think setting attributes outside the `__init__` is a no go, because it goes so strongly against what Python developers expect, and IDE linters will mark such as incorrect (both when writing amads code and when using amads as an external dependency). However, we could do something very similar which is to support dict-like indexing, so you can write for example `note["mobility"]` which will return a `KeyError` if the attribute is not found. To this, we'd just need to modify the `Event` definition as follows:

```py
class Event:
    def __init__(self, duration, delta):
        self.delta = delta
        self.duration = duration
        self._parent = None
        self._data = {}

    def __getitem__(self, key):
        return self._data[key]

    def __setitem__(self, key, value):
        self._data[key] = value
```

This is the best compromise I can currently think of.

amads/melody/gradus.py
MarkGotham: +1

amads/melody/mel_accent.py
MarkGotham: We could offer a `make_monophonic` function to map from "sort of" to "really" monophonic MIDI? 

Clear principles and documentation for the design needed.

E.g., each new note start takes precedence, cutting off everything else.

Then for any exactly coincident events? That would be one of the "user's fault" cases Peter mentions ;). Perhaps raising an error in that case is best, to draw a line under how far we're prepared to accommodate inappropriate scores (exact coincidence indicating a fully symbolic score in multiple parts).

amads/melody/meteraccent.py
```diff
@@ -0,0 +1,193 @@
...
+        Accent synchrony value between 0 and 1.
+        Higher values indicate greater synchronization between different
+        types of accents.
+        Returns 0 for empty scores or single notes.
+    
+    References
+    ----------
+    .. [1] Eerola, T., Himberg, T., Toiviainen, P., & Louhivuori, J. 
+           Perceived complexity of Western and African folk melodies by 
+           Western and African listeners.
```
MarkGotham, line 33: Year?

amads/melody/meteraccent.py
```diff
@@ -0,0 +1,193 @@
...
+        # Pad first note to match length
+        intervals = np.pad(intervals, (1, 0), 'constant')
+        # Normalize intervals to create accent weights
+        return intervals / np.max(intervals) if np.max(intervals) > 0 else np.zeros_like(intervals)
+    
+    def calculate_metrical_accents(notes: list) -> np.ndarray:
+        """Calculate metrical accents based on metric position."""
+        # Assuming 4/4 meter for simplicity
+        # Could be extended to handle different time signatures
+        METER_PATTERN = np.array([1.0, 0.2, 0.5, 0.2])  # 4/4 meter weights
```
MarkGotham, line 72: Let's call these `meter_weights` or similar please. Pattern is ambiguous.

MarkGotham: Worth moving this to an argument too. We will certainly offer different profiles.

Maybe also worth a note to say that this argument type may change.

Issue #64, opened 2025-02-12
Mutability


pmcharrison:
@rbdannenberg I was wondering if we should formalise our position on immutability? There are some things that are hard to implement efficiently if we can't assume that a `Score` is immutable. If we assume a Score is immutable, then certain properties can be computed once (e.g. ismonophonic) and then relied upon forever, instead of being potentially invalidated in the future. 

If we decide that Scores should be immutable, then we could consider enforcing this immutability by using dataclasses and setting `frozen=True`.

(cc. @Dav8Circle)

rbdannenberg:
That's a really good question. I think "pure" immutability is not a good idea: In particular, I found basic operations like flatten or stripping ties were best performed through incremental modifications which would not be possible or practical with purely immutable scores. On the other hand, I want to avoid situations like this: I load a score and perform analysis A followed by analysis B. But A modifies the score and changes what B sees. Very confusing and error-prone! So the goal should be that score modifications are not generally observable. That means it is fine to copy a score, mutate it to your heart's content, then return the copy. *Maybe* the mutations are even performed by calling other functions, but we should clearly document that these are "utility" functions that may modify the score and should be used carefully, and only on scores that are known not to be shared by others (e.g. you make a copy, call mutating utility functions and make local modification to the new score object). Comments?

rbdannenberg:
Another thing I've thought about is annotating scores with analysis data, e.g. add an attribute to each note with an estimate of entropy or tension or whatever. This is a mutation that does not necessarily impact any other uses of the same Score object. However, it's still safer to copy the Score before mutating, and that seems best.

pmcharrison:
That makes sense, that pure immutability might be too constraining. It's also something that is hard to backtrack from later if you change your mind.

If we can't assume immutability, then we need to think carefully about how to invalidate attributes upon changes to the score.

I am wondering about the following approach. We define derived attributes such as `is_monophonic` using a caching decorator, like this:

```py
class Score:
    version = 0

    def flag_modified(self):
        self.version += 1

    @cached
    def is_monophonic(self):
        for container in self.note_containers():
            notes = container.find_all(Note)
            if not _ismonophonic(notes):
                return False
        return True
```

We would implement a `cached` decorator stores the value of the method output alongside the current value of `self.version`. The next time the method is called, if `self.version` is unchanged it will return the previous output, otherwise it will rerun the method.

[...]

PR #65, opened 2025-02-13
Add M-Type Tokenizer

dmwhyatt:
Within the FANTASTIC toolbox, many features are calculated from M-Types, which are roughly equivalent to a musical n-gram. Under the schema FANTASTIC provides, an M-Type is a sequence of length `n`, comprised of intervals and ioi ratios, which appear in the present code as tuples. M-Types are used in FANTASTIC to calculate features analogous to those found in computational linguistics work, such as measures of lexical diversity or to compare target melodies to a corpus.

This PR adds a base `MelodyTokenizer` class, and `FantasticTokenizer` which computes M-Types from a `Score` object. The top level function `get_mtype_counts()` allows the user to specify the length of the M-Type computed, or whether they want all M-Types to be computed, which is the default option. 

Also added some tests and a demo.

### DIFF
--- a/amads/algorithms/mtype_tokenizer.py
+++ b/amads/algorithms/mtype_tokenizer.py
@@ -9,6 +9,10 @@ from amads.pitch.ismonophonic import ismonophonic
 class MelodyTokenizer:
     """Base class for tokenizing melodies into n-grams."""
 
+    def __init__(self):
+        """Initialize the tokenizer."""
+        self.ioi_data = {}  # Dictionary to store IOI information for notes
+
     def tokenize(self, score: Score) -> List[List]:
         """Tokenize a melody into phrases.
 
@@ -44,24 +48,31 @@ class MelodyTokenizer:
         flattened_score = score.flatten(collapse=True)
         notes = list(flattened_score.find_all(Note))
 
+        # Clear any previous IOI data
+        self.ioi_data = {}
+
         # Calculate IOIs and IOI ratios
         # n.b. when #68 is merged, this should be revised
         # Calculate IOIs
         for i, note in enumerate(notes):
-            if i > 1:
-                note.ioi = note.onset - notes[i - 1].onset
+            # Initialize entry for this note
+            self.ioi_data[note] = {"ioi": None, "ioi_ratio": None}
+
+            if i > 0:
+                self.ioi_data[note]["ioi"] = note.onset - notes[i - 1].onset
             else:
-                note.ioi = None
+                self.ioi_data[note]["ioi"] = None
 
             if i == 0:
-                note.ioi_ratio = None
+                self.ioi_data[note]["ioi_ratio"] = None
             else:
-                prev_ioi = notes[i - 1].ioi
-                ioi = note.ioi
+                prev_note = notes[i - 1]
+                prev_ioi = self.ioi_data[prev_note]["ioi"]
+                ioi = self.ioi_data[note]["ioi"]
                 if ioi is None or prev_ioi is None:
-                    note.ioi_ratio = None
+                    self.ioi_data[note]["ioi_ratio"] = None
                 else:
-                    note.ioi_ratio = ioi / prev_ioi
+                    self.ioi_data[note]["ioi_ratio"] = ioi / prev_ioi
 
         return notes
 
@@ -91,6 +102,7 @@ class FantasticTokenizer(MelodyTokenizer):
     """
 
     def __init__(self):
+        super().__init__()
         self.tokens = []
 
     def tokenize(self, score: Score) -> List:
@@ -116,10 +128,15 @@ class FantasticTokenizer(MelodyTokenizer):
 
         for prev_note, current_note in zip(notes[:-1], notes[1:]):
             pitch_interval = current_note.keynum - prev_note.keynum
-            if prev_note.ioi is None or current_note.ioi is None:
+            if (
+                self.ioi_data[prev_note]["ioi"] is None
+                or self.ioi_data[current_note]["ioi"] is None
+            ):
                 ioi_ratio = None
             else:
-                ioi_ratio = current_note.ioi / prev_note.ioi
+                ioi_ratio = (
+                    self.ioi_data[current_note]["ioi"] / self.ioi_data[prev_note]["ioi"]
+                )
 
             pitch_interval_class = self.classify_pitch_interval(pitch_interval)
             ioi_ratio_class = self.classify_ioi_ratio(ioi_ratio)
--- a/amads/melody/segment.py
+++ b/amads/melody/segment.py
@@ -29,13 +29,20 @@ def fantastic_segmenter(score: Score, phrase_gap: float, units: str) -> List[Sco
         # Extract notes from score
         flattened_score = score.flatten(collapse=True)
         notes = list(flattened_score.find_all(Note))
+
+        # Create a dictionary to store IOI information
+        ioi_data = {}
+
         # Calculate IOIs
         for i, note in enumerate(notes):
+            # Initialize entry for this note
+            ioi_data[note] = None
+
             # first note has no IOI by convention
             if i > 0:
-                note.ioi = note.onset - notes[i - 1].onset
+                ioi_data[note] = note.onset - notes[i - 1].onset
             else:
-                note.ioi = None
+                ioi_data[note] = None
 
         phrases = []
         current_phrase = []
@@ -43,35 +50,37 @@ def fantastic_segmenter(score: Score, phrase_gap: float, units: str) -> List[Sco
             # Check whether we need to make a new phrase
             need_new_phrase = (
                 len(current_phrase) > 0
-                and note.ioi
+                and ioi_data[note]
                 is not None  # Check current note's IOI instead of previous note
-                and note.ioi > phrase_gap  # Default phrase gap of 1 quarter note
+                and ioi_data[note] > phrase_gap
             )
             if need_new_phrase:
                 # Create new score for the phrase
-                phrase_score = Score()
-                part = Part()
+                phrase_score = Score(onset=0, duration=None)
+                part = Part(
+                    parent=None, onset=0, duration=None
+                )  # parent=None is required
                 start_time = current_phrase[0].onset
                 # Adjust note timings relative to phrase start
                 for phrase_note in current_phrase:
-                    new_note = phrase_note.deep_copy()
+                    new_note = phrase_note.copy()
                     new_note.onset -= start_time
                     part.insert(new_note)
-                phrase_score.insert(part)
+                phrase_score.insert(part)  # This will set the parent
                 phrases.append(phrase_score)
                 current_phrase = []
             current_phrase.append(note)
 
         # Append final phrase
         if len(current_phrase) > 0:
-            phrase_score = Score()
-            part = Part()
+            phrase_score = Score(onset=0, duration=None)
+            part = Part(parent=None, onset=0, duration=None)  # parent=None is required
             start_time = current_phrase[0].onset
             for phrase_note in current_phrase:
-                new_note = phrase_note.deep_copy()
+                new_note = phrase_note.copy()
                 new_note.onset -= start_time
                 part.insert(new_note)
-            phrase_score.insert(part)
+            phrase_score.insert(part)  # This will set the parent
             phrases.append(phrase_score)
 
         return phrases


Dav8Circle:
TODO: Based on what happens with #64, I will need to change the way `ioi` and `ioi_ratio` are integrated with `Note` objects, as I would expect straight up setting a new attribute (as I currently do) wouldn't work.

Dav8Circle:
Have further added some simple features from FANTASTIC which are calculated from the MType dictionaries we get from the Tokenizer.

amads/algorithms/mtype_tokenizer.py
```diff
@@ -0,0 +1,412 @@
...
+        if ioi_ratio is None:
+            return None
+        elif ioi_ratio < 0.8118987:
+            return "q"
+        elif ioi_ratio < 1.4945858:
+            return "e"
+        else:
+            return "l"
+
+    def count_grams(
```
pmcharrison, line 382: We need to find a way of avoiding duplicating the ngram logic from above.

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
+from collections import Counter
```
pmcharrison, line 1: I still need to review this file

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+import numpy as np
+
+
+def yules_k(ngram_counts: dict[tuple, int]) -> float:
+    """Calculates mean Yule's K statistic over m-type n-grams.
+    Yule's K is a measure of the rate at which words are repeated in a given text.
+    In the context of music, it measures the rate at which m-types are repeated in a given melody,
+    according to the formula:
+
+    :math:`K = 1000 * (sum(V(m,N) * m²) - N) / (N * N)`
```
pmcharrison, line 12: instead of `m²` probably better to write `m^2`

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+    where:
+    - :math:`V(m,N)` is the number of m-types with frequency :math:`m` in a melody with :math:`N` m-tokens
```
pmcharrison, line 14: Presumably it doesn't technically have to be a melody, but could be a corpus of melodies?

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+        return 0.0
+
+    # Calculate sum(vm * m²) where vm is frequency of value m
+    vm_m2_sum = sum(freq * (count * count) for count, freq in freq_spec.items())
+
+    # Calculate K with scaling factor of 1000
+    k = 1000 * (vm_m2_sum - total_tokens) / (total_tokens * total_tokens)
+    k_values.append(k)
+
+    return float(np.mean(k_values)) if k_values else 0.0
```
pmcharrison, line 68: I'm not sure about silently returning 0.0 in an empty sequence -- maybe raise a ValueError instead?

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+    >>> simpsons_d(high_repetition_counts)  # Higher D indicates less diversity
+    0.022...
+    >>> # Sample melody with low repetition
+    >>> score = Score.from_melody([60, 61, 62, 68, 59, 71], durations=[0.25, 0.5, 0.75, 0.35, 0.15, 0.25])
+    >>> tokenizer = FantasticTokenizer()
+    >>> low_repetition_counts = tokenizer.get_mtype_counts(score, method="all")
+    >>> simpsons_d(low_repetition_counts)  # Lower D indicates more diversity
+    0.00952...
+    >>> # Empty input
+    >>> simpsons_d({})
```
pmcharrison, line 109: Maybe this should be a ValueError too?

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+    >>> sichels_s(low_repetition_counts)  # Lower S indicates fewer doubles
+    0.0714...
+    >>> # Empty input
+    >>> sichels_s({})
+    0.0
+    """
+    if not ngram_counts:
+        return 0.0
+
+    s_values = []
```
pmcharrison, line 177: Likewise, is `s_values` really necessary?

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+    doubles = sum(1 for count in ngram_counts.values() if count == 2)
+
+    # Get total_types (total number of unique n-grams)
+    total_types = len(ngram_counts)
+
+    if total_types == 0:
+        return 0.0
+
+    # Calculate S value
+    s = float(doubles) / total_types
```
pmcharrison, line 189: I feel like `|n|` has been missed from this formula

tests/test_mtype_tokenizer.py
```diff
@@ -0,0 +1,92 @@
...
+    # The first phrase has 7 notes in so the maximum length MType is 6
+    assert len(tokenizer.phrases[0]) == 7
+    with pytest.raises(ValueError):
+        tokenizer.get_mtype_counts(melody_2, method=7)
+
+    # This is true of the second phrase as well
+    assert len(tokenizer.phrases[1]) == 7
+    with pytest.raises(ValueError):
+        tokenizer.get_mtype_counts(melody_2, method=7)
+
```
pmcharrison, line 89: Somewhere in here we need a basic test of `get_ngram_counts` which makes sure that eg.

```py
[0, 1, 1, 0, 1]
```

gets exactly the right n-gram counts

demos/mtype_tokenizer.py
```diff
@@ -0,0 +1,44 @@
+from amads.algorithms.mtype_tokenizer import FantasticTokenizer
```
pmcharrison, line 1: Idea for API:

```py
tokenizer = MTypeTokenizer()
ngrams = NGramCounter()

for melody in corpus:
    phrases = segment_melody(melody)
    for phrase in phrases:
        tokens = tokenizer.tokenize(phrase)
        ngrams.count(tokens)
    
ngrams.honores_h
ngrams.productivity
```

amads/algorithms/mtype_tokenizer.py
```diff
@@ -0,0 +1,412 @@
+import warnings
+from collections.abc import Hashable
```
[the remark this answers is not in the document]
Dav8Circle: `collections.abc` is the new permanent residence for `Hashable`, as it's deprecated in `typing` and `collections`!

amads/algorithms/mtype_tokenizer.py
```diff
@@ -0,0 +1,412 @@
...
+
+    Attributes
+    ----------
+    precision : int
+        Number of decimal places to round IOI values to
+    phrases : list
+        List of melody phrases after segmentation
+    """
+
+    def __init__(self):
```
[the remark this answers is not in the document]
Dav8Circle: done!

amads/algorithms/mtype_tokenizer.py
```diff
@@ -0,0 +1,412 @@
...
+            If "all", count n-grams of all lengths.
+            If int, count n-grams of that specific length, by default "all"
+
+        Returns
+        -------
+        dict
+            Counts of each unique n-gram
+        """
+        self.tokenize_melody(score)
+        return self.ngram_counts(method)
```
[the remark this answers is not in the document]
Dav8Circle: Good shout, moved

amads/algorithms/mtype_tokenizer.py
```diff
@@ -0,0 +1,412 @@
...
+        -------
+        list[Note]
+            List of notes with IOI and IOI ratio values calculated
+        """
+        flattened_score = score.flatten(collapse=True)
+        notes = list(flattened_score.find_all(Note))
+
+        # Calculate IOIs and IOI ratios
+        # n.b. when #68 is merged, this should be revised
+        for i, note in enumerate(notes):
```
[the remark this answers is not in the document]
Dav8Circle: Having looked through Fantastic, they seem to use forwards, so I guess the last note would have undefined ioi - do we want to bring it in line with common practice, or match implementation?

amads/complexity/mtype_features.py
Dav8Circle: Good point, well made

amads/algorithms/mtype_tokenizer.py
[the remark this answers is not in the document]
pmcharrison: I see, so what was this statement for? 

> collections.abc is currently not supported in Python 3.13.1

amads/algorithms/mtype_tokenizer.py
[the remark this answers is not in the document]
pmcharrison: Thinking about it, I think we have to define it with respect to the previous note. Otherwise it is weirdly inconsistent with pitch interval. I think also it would be helpful to be consistent with IDyOM given that an implementation is coming soon.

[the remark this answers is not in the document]
rbdannenberg: I'm adding an Issue about intervals and their representation.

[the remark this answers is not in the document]
pmcharrison: Was thinking about pinging you ;)

amads/algorithms/mtype_tokenizer.py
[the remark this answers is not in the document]
Dav8Circle: I was using a 3.13.1 venv and couldn't work out why `collections.abc` wasn't working... turns out it's a known issue as far as I can tell: https://github.com/pylint-dev/pylint/issues/10112

Ought to be resolved apparently, but not yet for me!

demos/mtype_tokenizer.py
Dav8Circle: How would you feel about a slightly more in-built functionality? e.g. a higher level function `get_mtype_counts()` in `NGramCounter()`. Something like this, inside of the `NGramCounter` class:

```py
def get_mtype_counts(self, method: Union[str, int] = "all") -> Dict:
    tokenizer = FantasticTokenizer()
    tokenizer.tokenize_melody(self.score)
    self.ngram_counts = self._ngram_counts(tokenizer.tokens, method)
    return self.ngram_counts
```

This I think just simplifies the user experience a bit: instead of initialising both tokenizer and counter, having the user select the function they want, which then decides which tokenizer should be used

amads/complexity/mtype_features.py
Dav8Circle: Fixed

amads/complexity/mtype_features.py
Dav8Circle: I realised yesterday that it was missing from all of them - have updated now

demos/mtype_tokenizer.py
Dav8Circle: Actually, this makes it harder to access `phrase_gap` as I have now noticed, so will probably not do this

Dav8Circle:
Have enacted the larger architectural changes, still got a lot of the nitty gritty detail/docs to iron out!

amads/complexity/mtype_features.py
Dav8Circle: Yeah, I think you're right, we want people to know that the calculation has failed, rather than thinking that 0.0 is a real calculated value

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+    # Calculate K with scaling factor of 1000
+    k = 1000 * (vm_m2_sum - total_tokens) / (total_tokens * total_tokens)
+    k_values.append(k)
+
+    return float(np.mean(k_values)) if k_values else 0.0
+
+
+def simpsons_d(ngram_counts: dict[tuple, int]) -> float:
+    """Compute mean Simpson's D diversity index over m-type n-grams.
+    Simpson's D is also a measure of the diversity of m-types in a melody,
```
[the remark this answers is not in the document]
Dav8Circle: Sure thing!

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
+from collections import Counter
+
+import numpy as np
```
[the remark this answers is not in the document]
Dav8Circle: I agree, and actually, this means we need to generalise the docs for the functions too - these features are calculable for all ngram tokens, so the docs need to reflect that!

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+
+    # Get hapax_count (number of hapax legomena)
+    hapax_count = sum(1 for count in ngram_counts.values() if count == 1)
+
+    # Get total_types (total types)
+    total_types = len(ngram_counts)
+
+    # Handle edge cases
+    if total_types == 0 or hapax_count == 0 or hapax_count == total_types:
+        return 0.0
```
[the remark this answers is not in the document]
Dav8Circle: updated with (hopefully) informative error messages

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+    >>> honores_h(low_repetition_counts)  # Higher H indicates more unique MTypes
+    3325.68...
+    >>> # Empty input
+    >>> honores_h({})
+    0.0
+    """
+    if not ngram_counts:
+        return 0.0
+
+    h_values = []
```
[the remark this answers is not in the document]
Dav8Circle: no :(

amads/complexity/mtype_features.py
pmcharrison: Done

amads/algorithms/mtype_tokenizer.py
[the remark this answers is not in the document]
pmcharrison: Is it just a pylint problem? The code still works fine when run?

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+
+def mean_entropy(ngram_counts: dict[tuple, int]) -> float:
+    """Compute entropy of m-type n-gram distribution.
+    Calulates the mean Shannon entropy of the m-type n-gram distribution.
+    This is defined as:
+    :math:`H = -sum(p(x) * log2(p(x)))`
+    where:
+    - :math:`p(x)` is the probability of the i-th m-type
+    - :math:`N` is the total number of m-types in the melody
+    - :math:`H` is the Shannon entropy
```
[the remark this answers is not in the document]
Dav8Circle: yeah that is a lot clearer, sorry!

amads/complexity/mtype_features.py
```diff
@@ -0,0 +1,386 @@
...
+    entropy_norm = entropy / np.log2(total_tokens)
+    entropy_values.append(entropy_norm)
+
+    return float(np.mean(entropy_values)) if entropy_values else 0.0
+
+
+def mean_productivity(ngram_counts: dict[tuple, int]) -> float:
+    """Compute mean productivity of m-type n-gram distribution.
+
+    Mean productivity is defined as the mean over all lengths of the number of m-types
```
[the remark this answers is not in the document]
Dav8Circle: yeah that doesn't make sense, my bad

demos/mtype_tokenizer.py
```diff
@@ -0,0 +1,44 @@
...
+    print(f"Dictionary of bigram counts: {bigram_counts}\n")
+
+    # Count only trigrams (n=3)
+    trigram_counts = tokenizer.get_mtype_counts(melody_1, method=3)
+    print(f"Dictionary of trigram counts: {trigram_counts}\n")
+
+    # Count all n-grams
+    all_counts = tokenizer.get_mtype_counts(melody_1, method="all")
+    print(f"Dictionary of all n-gram counts: {all_counts}\n")
+
```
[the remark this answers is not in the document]
Dav8Circle: still TODO

demos/mtype_tokenizer.py
pmcharrison: Yes; I think ideally you want to avoid mixing concerns like this. ngram counting is a general statistical process that ideally should not have anything melody-specific based into it.

amads/algorithms/mtype_tokenizer.py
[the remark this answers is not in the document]
pmcharrison: Great!

amads/algorithms/ngrams.py
```diff
@@ -0,0 +1,329 @@
...
+        self.ngram_counts = None  # Initialize ngram_counts as None
+
+    def count_ngrams(self, tokens: list, method: Union[str, int]) -> Dict:
+        """Count n-grams in a sequence of tokens.
+
+        Parameters
+        ----------
+        tokens : list
+            List of tokens to process (will be flattened if nested)
+        method : str or int
```
pmcharrison, line 19: I think better to name method 'n', and for the option to be either None or int

amads/algorithms/ngrams.py
```diff
@@ -0,0 +1,329 @@
...
+            If "all", count n-grams of all possible lengths.
+            If int, count n-grams of that specific length only.
+
+        Returns
+        -------
+        dict
+            Dictionary mapping each n-gram to its count
+        """
+        # Flatten nested list structure
+        flat_tokens = (
```
pmcharrison, line 29: I don't think this is a good idea, because then ngrams will be created by merging the end of one sequence with the beginning of the other. I would just force `tokens` to be a list not a nested list

amads/algorithms/ngrams.py
```diff
@@ -0,0 +1,329 @@
...
+            else tokens
+        )
+
+        if not flat_tokens:
+            return {}
+
+        if method == "all":
+            n_values = range(1, len(flat_tokens) + 1)
+        else:
+            if not isinstance(method, int):
```
pmcharrison, line 41: Presumably you should allow people to specify a list of ns rather than just a single n?

pmcharrison (commented):
Some comments on ngrams

amads/algorithms/ngrams.py
pmcharrison: ie if people want to iterate over a corpus, they should write `for sequence in corpus`

Dav8Circle: Ah yeah, this was a bodgy fix - the nested list was produced by a bad list comp, fixed now!

amads/algorithms/ngrams.py
Dav8Circle: Done

amads/algorithms/mtype_tokenizer.py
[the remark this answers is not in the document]
Dav8Circle: done now

amads/melody/segment.py
```diff
@@ -0,0 +1,48 @@
...
+    list[Note]
+        List of Note objects representing phrases
+    """
+    # Extract notes from score
+    flattened_score = score.flatten(collapse=True)
+    notes = list(flattened_score.find_all(Note))
+    # Calculate IOIs
+    for i, note in enumerate(notes):
+        if i < len(notes) - 1:
+            note.ioi = notes[i + 1].start - note.start
```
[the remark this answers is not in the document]
Dav8Circle: wait, #67 isn't merged yet! will do then

amads/melody/segment.py
```diff
@@ -0,0 +1,48 @@
...
+        else:
```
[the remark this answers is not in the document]
Dav8Circle: done

amads/melody/segment.py
```diff
@@ -0,0 +1,48 @@
+from typing import List
+
+from amads.core.basics import Note, Score
+
+
+def fantastic_segmenter(score: Score, phrase_gap: float = 1.0) -> List[Note]:
```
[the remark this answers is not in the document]
Dav8Circle: All done

amads/melody/segment.py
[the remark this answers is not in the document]
pmcharrison: Merged now

amads/algorithms/mtype_tokenizer.py
```diff
@@ -0,0 +1,347 @@
...
+        Number of decimal places to round IOI values to
+    phrases : list
+        List of melody phrases after segmentation
+    """
+
+    def __init__(self, precision: int = 6):
+        self.precision = precision
+        self.phrases = []
+
+    def tokenize_melody(self, score: Score) -> List[List]:
```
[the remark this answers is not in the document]
Dav8Circle: done, and with extra function to get all complexity features as a dict, for ease of use

demos/mtype_tokenizer.py
```diff
@@ -27,7 +29,7 @@ def example_usage():
     ngrams = NGramCounter()
 
     # Tokenize melody and count n-grams
-    tokenizer.tokenize_melody(melody)
+    tokenizer.tokenize(melody)
```
pmcharrison, line 32: Let's instead return the tokens from the tokenizer, like this:

```py
tokens = tokenizer.tokenize(melody)
```

pmcharrison: (and don't store it in the tokenizer)

demos/mtype_tokenizer.py
```diff
@@ -98,9 +100,19 @@ def example_usage():
...
+    for segment in segments:
+        segment_tokens = tokenizer.tokenize(segment)
+        all_tokens.extend(segment_tokens)
 
-    tokenizer.tokenize_melody(happy_birthday)
-    ngrams.count_ngrams(tokenizer.tokens, n=None)
+    # Initialize NGramCounter
+    ngrams = NGramCounter()
+    # Count all n-grams
+    all_counts = ngrams.count_ngrams(all_tokens, n=None)
```
pmcharrison, line 115: Don't want ngrams to span phrases. Instead:

```py
    segments = fantastic_segmenter(happy_birthday, phrase_gap=0.75, units="quarters")
    # Tokenize each segment and collect all tokens
    tokenizer = FantasticTokenizer()
    ngrams = NGramCounter()

    for segment in segments:
        segment_tokens = tokenizer.tokenize(segment)
        ngrams.count_ngrams(segment_tokens)

    print(f"Yule's K: {ngrams.yules_k}")
```

demos/mtype_tokenizer.py
```diff
@@ -98,9 +100,19 @@ def example_usage():
...
     # These are the complexity measures computed from the n-grams of all lengths
     print("All n-grams:")
```
pmcharrison, line 117: Need to update `NGramCounter` such that it is stateful. `count_ngrams` shouldn't return anything. It should not reset the counter each time, rather the counter should be initialised in `__init__`, then each time you call `count_ngrams`, it should update the ngram counter from the current sequence. That way you can process many separate sequences with the same ngram counter.

demos/mtype_tokenizer.py
Dav8Circle: done

demos/mtype_tokenizer.py
Dav8Circle: Think I've addressed this with the newest version

demos/mtype_tokenizer.py
Dav8Circle: Got it now, thanks

pmcharrison (approved):

tests/test_mtype_tokenizer.py
```diff
@@ -0,0 +1,188 @@
...
+            0.5,
+            0.5,
+            0.5,
+        ],
+    )
+
+    # Test that the melody is segmented into 2 phrases
+    segments = fantastic_segmenter(melody_2, phrase_gap=0.75, units="quarters")
+    # Extract notes from each segment for counting
+    segment_notes = [list(segment.find_all(Note)) for segment in segments]
```
pmcharrison, line 95: This has the same problem we talked about before, where ngrams are going to formed across segment boundaries.

amads/algorithms/mtype_tokenizer.py
```diff
@@ -0,0 +1,303 @@
...
+    [1] Müllensiefen, D. (2009). Fantastic: Feature ANalysis Technology Accessing
+        STatistics (In a Corpus): Technical Report v1.5
+    """
+
+    def __init__(self):
+        """Initialize the tokenizer.
+
+        Parameters
+        ----------
+        phrase_gap : float, optional
```
pmcharrison, line 109: These parameters need to be removed from the docstring

amads/algorithms/mtype_tokenizer.py
```diff
@@ -0,0 +1,303 @@
...
+
+from amads.core.basics import Note, Score
+
+
+class MelodyTokenizer:
+    """Base class for tokenizing melodies into n-grams.
+
+    Attributes
+    ----------
+    phrases : list
```
pmcharrison, line 13: Seems not relevant now?

amads/algorithms/mtype_tokenizer.py
```diff
@@ -0,0 +1,303 @@
...
+
+    An M-Type is a sequence of musical symbols (pitch intervals and duration ratios)
+    that represents a melodic fragment, similar to how an n-gram represents a sequence
+    of n consecutive items from a text. The length of an M-Type can vary, just like
+    n-grams can be of different lengths (bigrams, trigrams, etc.)
+
+    The tokenizer takes a score as the input, and returns a dictionary of unique
+    M-Type (n-gram) counts.
+
+    Parameters
```
pmcharrison, line 84: Are these parameters `phrase_gap` and `units` still relevant? Maybe we should remove them?

amads/algorithms/mtype_tokenizer.py
Dav8Circle: done

amads/algorithms/mtype_tokenizer.py
Dav8Circle: removed

amads/algorithms/mtype_tokenizer.py
Dav8Circle: Done!

Issue #45, opened 2024-12-15
Questioning weakref use in parent


pmcharrison:
We have this:

```py
    @parent.setter
    def parent(self, p):
        self._parent = weakref.ref(p)
```

As I understand it, saving `parent` as a weakref means that Python will garbage-collect the parent if the parent is not saved separately as a variable. So, in this code:

```py
note = read_midi("...").notes[0]
```

the `note.parent` attribute will potentially be unavailable because it the parent might have been garbage collected. Is this what we want? Maybe @rbdannenberg can comment?

rbdannenberg:
I was under the impression the Python does not reclaim structures with cycles due to reference counting. But in learning about weak references, it seems there is some kind of tracing garbage collection in addition to reference counting, but it's unclear if relying on that is a reasonable thing, given all the warnings about cycles and reference counting. If cycles can actually be reclaimed (and there's no problem with it -- it might be expected to occur a lot), then I think we should make parent an ordinary reference. Good question! I would strongly recommend against keeping a reference to a note if there's no reference to the parents. If the note has some use outside of the parents, it should probably be copied into another score structure, freeing the original note.

pmcharrison:
Yes, I do think Python has a cycle detector, see the last paragraph of [this page](https://docs.python.it/html/ext/refcounts.html). Though it is possible to configure a Python installation with the cycle detector I think this is a perverse use case... So probably it is best just to not worry and rely on the built-in cycle detector?

rbdannenberg:
I think you mean it is possible to configure Python _without_ the cycle detector, but this is a perverse case. I agree. So I now think it's best to eliminate the weak reference to parent and just allow the cycles it will create. If you create scores without cycle detection, they will not be freed automatically, so you'll leak memory, but even then, it's unlikely anyone will notice given the size of scores. Now I'm curious about how much memory Python objects take -- it would be interesting to know the approximate size of a score with 1000 notes just for back-of-the-envelope calculations.

pmcharrison:
Sorry, you're right, I meant without! I agree, it would be useful to know how much memory the objects take. Will see if I can find that out sometime.

diwangislucky:
I was under the impression that, while cycles are able to be handled by the garbage collection, it takes longer for the garbage collector to detect and free cyclic references. This is why the discussion I've seen on this topic seems to allow weak references as an acceptable compromise to break cyclic references. However, given the size of the scores, not sure if it's worth it worrying over a longer memory lifecycle?

[...]

PR #68, opened 2025-02-13
Implementing caching of certain Score-level attributes; implement several new Note attributes

pmcharrison:
This PR implements caching of certain Score-level attributes. The caches are invalidated when the score is changed.

As part of testing this functionality, the PR also implements `Score.is_monophonic`, `Score.notes`, `note.previous_note`, `note.next_note`, `note.ioi`, and `note.ioi_ratio`.

The caching is useful for `note.next_note` and `note.previous_note` because it enables us to maintain a lazily computed dictionaries that link each note with its predecessors and successors. It is likewise useful for `note.ioi` because it allows `note.ioi` to refer to a global score attribute (a dictionary logging onset times) without requiring this attribute to be recomputed for each note.

In order to resolve a blocking bug, the PR also implements #45, removing the use of weakref in `Event.parent`.


pmcharrison:
@Dav8Circle I was prompted to do this by your draft PR that was wondering how `ioi` should be implemented. I think this provides quite an effective solution.

pmcharrison:
PS the PR also provides a `Score.notes` attribute which helps you out too.

rbdannenberg (commented):
This is a really interesting use of meta-programming (if that's the right term for this).  However, I'm worried about the computational cost. Isn't *every* attribute going to invalidate *every* value in the caches from the Event up to the Score level? I think things are going to be pretty slow to begin with, so to turn every assignment into a chain of method calls and allocations (to replace the caches) could slow things down by an order of magnitude.

pmcharrison:
Thanks, @rbdannenberg. You are right that the cache invalidation is quite liberal, i.e. changing one note's attribute will invalidate multiple attributes across multiple notes in the score. However, I think the computational cost is low, because of the way that cache inheritance works. Specifically, when I set the attribute on a note, the following operations are triggered:

- Invalidate cache on parent Chord (if exists)
- Invalidate cache on parent Measure (if exists)
- Invalidate cache on parent Staff (if exists)
- Invalidate cache on parent Part (if exists)
- Invalidate cache on parent Score

Accessing the parent is a simple attribute lookup and invalidating the cache is a very cheap operation (creating an empty dictionary if it's not already empty). Note that we do not need to iterate over all notes in the score, we just need to chase the hierarchy up to a maximum of 5 parents. This should be a relatively cheap process in the context of everything else we are doing.

I admit that it is quite possible that this change would making setting note attributes slower by an order of magnitude, but attribute setting is one of the fastest operations you can do in Python, so I'm not sure it would slow us down in any noticeable way?

PR #70, opened 2025-02-16
converted parent to be a regular attribute and not a weak link

rbdannenberg:
Based on earlier discussions, I changed parent back to a normal reference, depending on tracing GC to free cycles that result. There seem to be no uses of parent outside of core/basics.py.

### DIFF
--- a/amads/core/basics.py
+++ b/amads/core/basics.py
@@ -33,12 +33,11 @@ Score (one per musical work or movement)
 
 
 import functools
-import weakref
 from math import floor
 from typing import List, Optional, Union
 
-from .time_map import TimeMap
 
+from .time_map import TimeMap
 
 class Event:
     """Event is a superclass for Note, Rest, EventGroup, and just about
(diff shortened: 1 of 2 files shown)


pmcharrison:
Thanks @rbdannenberg! Now that you've installed the precommit you can run `pre-commit run --all-files`, commit the resulting (small) changes, then push again, and the tests should pass.

rbdannenberg:
Should pre-commit be run every time you commit?

rbdannenberg:
I was afraid of that -- in that case, here's what happened: I added the note on installing pre-commit to contributing.rst, and committed again, but tests still failed.

pmcharrison:
By default it only runs on the files you just changed, though, so it would not fix the problems in the previous commits. That's why you have to run `pre-commit run --all-files`

rbdannenberg:
OK, so I'm editing contributing.rst to:
```
GitHub contribution workflow
---------------------------

0. Use pre-commit
~~~~~~~~~~~~~~~~~
1. Install pre-commit hooks::

   pip install pre-commit && pre-commit install                                                                                                                           

2. Run pre-commit::

   pre-commit run --all-files                                                                                                                                             

(You only need to do this once after you have installed pre-commit.)
```

pmcharrison:
Thanks!

PR #71, opened 2025-02-17
Created music21_midi_import.py

sksound:
This function imports a MIDI file and converts it into a Music21 score object, extracting individual parts and notes. It does this by flattening the structure for easier access and optionally printing the textual representation of the musical content.

### DIFF
--- a/amads/io/music21_midi_import.py
+++ b/amads/io/music21_midi_import.py
@@ -1,0 +1,39 @@
+from music21 import converter, stream, note, chord
+
+def music21_midi_import(filename, m21print=False):
+    """
+    Import a MIDI file using Music21.
+    
+    Parameters:
+    - filename (str): Path to the MIDI file.
+    - m21print (bool): If True, prints the musical parts in textual format.
+    
+    Returns:
+    - score (music21.stream.Score): A Music21 score object containing the imported parts.
+    """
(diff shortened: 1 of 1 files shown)


amads/io/music21_midi_import.py
```diff
@@ -0,0 +1,39 @@
+from music21 import converter, stream, note, chord
+
+def music21_midi_import(filename, m21print=False):
+    """
+    Import a MIDI file using Music21.
+    
+    Parameters:
+    - filename (str): Path to the MIDI file.
+    - m21print (bool): If True, prints the musical parts in textual format.
```
rbdannenberg, line 9: What are "musical parts"? (Note that Std. MIDI Files have tracks and channels.)

amads/io/music21_midi_import.py
```diff
@@ -0,0 +1,39 @@
...
+    
+    Returns:
+    - score (music21.stream.Score): A Music21 score object containing the imported parts.
```
rbdannenberg, line 12: I was expecting the return value to be an AMADS Score object. I want to implement input/output with Music21 as well as with Partitura, and it could be useful to have a way to import from Music21 score to AMADS Score, but we don't use Music21 scores anywhere, so there's no need to convert MIDI to Music21.

Issue #67, opened 2025-02-13
onset/offset vs start/end


pmcharrison:
Currently we have `note.start` and `note.end`. How do we feel about this? do we prefer it to `note.onset` and `note.offset`? The latter is maybe confusing because originally `note.offset` was being used to mean something else, but that's changed now. The good thing about `onset` and `offset` is that it mirrors usage in MIDI and also (I think) in academic writing. Any thoughts @rbdannenberg ?

pmcharrison:
The reason I'm thinking about this is I'm in the process of writing `inter_onset_interval` which is definitely more standard that `inter_start_interval`.

rbdannenberg:
Yes, I agree we should be consistent and avoid two terms (onset and start) for the same concept. MIDI uses note-on and note-off messages, but I'm unaware of any use of onset/offset in MIDI. In fact, if you Google MIDI offset, the most common usage refers to latency between the message time and the note onset time, e.g. a sequencer can have an adjustable track offset, which has nothing to do with our use of offset. (But I still agree to eliminate start and stop when onset and offset can be used instead.)

rbdannenberg:
I wouldn't object to using ioi instead of inter_onset_interval -- I think IOI is so useful and important perceptually that 7-syllables is too many and the acronym is justified.

pmcharrison:
Sorry about the confusion re MIDI! OK, let's adopt onset/offset then. Good idea to shorten to ioi, I'll do that then.

PR #72, opened 2025-02-17
Rename .start -> .onset and .end -> .offset

pmcharrison:
Implements #67


pmcharrison:
@rbdannenberg hope this looks ok?

rbdannenberg:
Oops - I started renaming as well - but happy to see it done. I'm traveling for a few days, so I'll have to pick up again on Wed or Thurs to review. BTW, I'm looking at copy and deepcopy.

pmcharrison:
Thanks!

amads/algorithms/scale.py
```diff
@@ -27,17 +27,17 @@
...
-        scale(score, factor, "start")
+        scale(score, factor, "onset")
         return score
     for elem in score.content:
         if isinstance(elem, EventGroup):
             scale(elem, factor, dim)
-            if dim == "start":
-                elem.start *= 2
+            if dim == "onset":
+                elem.onset *= 2
```
rbdannenberg, line 36: I thought I saw somewhere this was changed to elem.onset *= factor (and it should be)

amads/algorithms/scale.py
```diff
@@ -2,21 +2,21 @@
...
     - 'duration': scales the durations of ALL NON-EVENTGROUP EVENTS (Note, Rest)
-    - 'all': scales both the start and the duration
+    - 'all': scales both the onset and the duration
 
 Output:
 SCORE = the scaled version of the input object
 
 Examples:
-scaled_score = scale(score.copy(),'start',2); % scales start by a factor of 2
+scaled_score = scale(score.copy(),'onset',2); % scales onset by a factor of 2
```
rbdannenberg, line 19: Shouldn't this make a deep copy? (I'm in the process of changing copy/deepcopy in another branch, so I'm fine to leave this as is and fix it elsewhere, but can we specify that the implementation will not modify score?)

amads/algorithms/scale.py
```diff
@@ -27,17 +27,17 @@
...
-                elem.start *= 2
+            if dim == "onset":
+                elem.onset *= 2
         else:
             if dim == "duration":
                 elem.duration *= factor
             elif dim == "start":
-                elem.start *= 2
-        score.duration = max(score.duration, elem.end)
+                elem.onset *= 2
```
rbdannenberg, line 41: elem.onset *= factor?

rbdannenberg (commented):
The score modification in scale.py is orthogonal to renaming start/end, so feel free to commit this (see also other comments) and we can address the mutation issue separately.

amads/algorithms/scale.py
pmcharrison: Yes, you're right. I'll change it here too just to be safe.

amads/algorithms/scale.py
pmcharrison: Yep

PR #80, opened 2025-03-06
Create norm.py

MarkGotham:
I'm flexible on the details of the implementation; simply getting the ball rolling to have this kind of shared functionality somewhere central (cf Huw's entropy PR).

### DIFF
--- a/amads/algorithms/norm.py
+++ b/amads/algorithms/norm.py
@@ -1,0 +1,134 @@
+"""
+Local implementation of standard normalization routines centrally for use in various functions.
+The local implementation serves to obviate the need for external libraries (e.g., scipy) in basic cases.
+
+Comparisons on this modules support any pair of profiles.
+Example use cased include pitch class profile matching for 'best key' measurement,
+and the metrical equivalent.
+"""
+
+import numpy as np
+
+__author__ = "Mark Gotham"
+
+
+# ------------------------------------------------------------------------------
+
+l1_names = ["sum", "manhattan", "city", "cityblock" "taxicab", "l1"]
+l2_names = ["euclidean", "l2"]
+max_names = ["max", "maximum", "inf", "infinity"]
+
+
+def normalize(
+    profile: list,
+    method: str = "Euclidean",
+    round_output: bool = False,
+    round_places: int = 3,
+) -> np.array:
+    """
+    Normalize usage profiles in standard ways.
+
+    Parameters
+    -------
+    profile: list
+        Any 1-D list of numeric data.
+
+    method: str
+        The desired normalization routine. Chose from any of
+        'Euclidean' or 'l2' ();
+        'Sum', 'Manhattan', or 'l1' (divide each value by the total across the profile);
+        'max', 'maximum', 'inf' or 'infinity' (divide each value by the largest value).
+        These strings are not case-sensitive (unaffected by presence/absence of initial caps etc).
+
+    round_output: bool
+        Optionally, round the output values (default False).
+
+    round_places: int
+        If rounding, how many decimal places to use (default=3). Moot if `round_output` if False (default).
+
+    Returns
+    -------
+    np.array
+
+    Examples
+    --------
+    >>> normalize([0, 1, 2, 3, 4], method="l1")
+    array([0. , 0.1, 0.2, 0.3, 0.4])
+
+    >>> normalize([0, 1, 2, 3, 4], method="l2")
+    array([0.        , 0.18257419, 0.36514837, 0.54772256, 0.73029674])
+    """
+
+    if np.max(profile) == 0:
+        return profile  # All 0s: don't divide by 0 (or indeed do anything!)
+
+    method = method.lower()
+    if method in l1_names:
+        norm_ord = 1
+    elif method in l2_names:
+        norm_ord = 2
+    elif method in max_names:
+        norm_ord = np.inf
+    else:
+        raise ValueError(
+            f"Invalid method. Must be one of {l1_names + l2_names + max_names}"
+        )
+    norm_dist = profile / np.linalg.norm(profile, ord=norm_ord)
+
+    if round_output:
+        return list(np.round(norm_dist, round_places))
+    else:
+        return norm_dist
+
+
+def shared_length(profile_1: list, profile_2: list) -> int:
+    """
+    Simple checks that two lists are of the same length.
+    If so, returns the length of the list; if not, raises an error.
+
+    >>> shared_length([1, 2], [2, 1])
+    2
+    """
+    ln1 = len(profile_1)
+    ln2 = len(profile_2)
+    if ln2 != ln1:
+        raise ValueError(f"Lengths (currently {ln1} and {ln2}) must match")
+    else:
+        return ln1
+
+
+def manhattan_distance(profile_1: list, profile_2: list) -> float:
+    """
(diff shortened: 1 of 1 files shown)


amads/algorithms/norm.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    ln1 = len(profile_1)
+    ln2 = len(profile_2)
+    if ln2 != ln1:
+        raise ValueError(f"Lengths (currently {ln1} and {ln2}) must match")
+    else:
+        return ln1
+
+
+def manhattan_distance(profile_1: list, profile_2: list) -> float:
```
rbdannenberg, line 108: This is not the standard definition, which does not involve normalization. Same for euclidean_distance.

amads/algorithms/norm.py
```diff
@@ -0,0 +1,161 @@
...
+    Normalize usage profiles in standard ways.
+
+    Parameters
+    -------
+    profile: list
+        Any 1-D list of numeric data.
+
+    method: str
+        The desired normalization routine. Chose from any of
+        'Euclidean' or 'l2' ();
```
rbdannenberg, line 38: 'Euclidean' or 'l2' (scale so the vector length is 1)

amads/algorithms/norm.py
pmcharrison: Yes, perhaps there should be a `normalize` parameter which defaults to False

amads/algorithms/norm.py
```diff
@@ -0,0 +1,161 @@
...
+    >>> manhattan_distance(profile_1, profile_2)
+    0.2
+    """
+    shared_length(profile_1, profile_2)
+    profile_1 = normalize(profile_1, "l1")
+    profile_2 = normalize(profile_2, "l1")
+    return float(sum([abs(profile_1[n] - profile_2[n]) for n in range(len(profile_1))]))
+
+
+def euclidean_distance(profile_1: list, profile_2: list) -> float:
```
pmcharrison, line 128: You could consider implementing pnorm_distance and then Manhatten and Euclidean distances are just special cases

amads/algorithms/norm.py
MarkGotham: Thanks. Done.

MarkGotham:
Hi @pmcharrison , @rbdannenberg ,

Ok, so, here's an update including the pnorm alternative. We wouldn't keep both options. They're here for comparison.

Results of a quick `timeit` experiment:

```
% python -m timeit -s "import norm" "norm.euclidean_distance([0, 1, 2], [2, 3, 4])"  
500000 loops, best of 5: 690 nsec per loop

% python -m timeit -s "import norm" "norm.pnorm_distance([0, 1, 2], [2, 3, 4])"
100000 loops, best of 5: 2.77 usec per loop

% python -m timeit -s "from scipy.spatial import distance" "distance.euclidean([0, 1, 2], [2, 3, 4])"
100000 loops, best of 5: 2.01 usec per loop
```

So that suggest that the local implementation of `euclidean_distance` is fastest by a factor of about 3x.

To review
- The case for scipy:
    - Easiest to create and maintain (already there).
- The case for local implementations
    - May be faster;
    - No user installation (scipy not bundled as standard).
    - Flexibility on our side to adapt arguments etc as needed.

I don't feel strongly either way. Simply
- another case for a repo-wide decision.
- if we do them locally then store centrally.

amads/algorithms/norm.py
pmcharrison: My thought was then once you have `pnorm_distance` you can then rewrite `euclidean_distance` to just call `pnorm_distance` with `p=2`?

MarkGotham: Can do , but see the `timeit` results. Can you replicate?

MarkGotham:
> I would agree that it makes sense to import scipy for a complex function like `get_peaks`, but these distance functions are one-liners so I don't think there is any real maintenance cost on our end!

Ok, so we do this (and more) set of norms and distances etc? Sounds good.

Then there's the question of how to implement. Can you replicate the `timeit` result? Or do you see another basis for deciding between them?

pmcharrison:
I think we shouldn't worry too much about the speed differences for now, that feels like [premature optimisation](https://wiki.c2.com/?PrematureOptimization) to me! Your implementation of Euclidean is quite different to pnorm; the former uses a list comprehension, the latter uses numpy instead. numpy is normally faster than what we can do in Python, but that typically only bears out once you have longer vectors.

MarkGotham:
Ok, fair enough. Alternatively it can be simply a way to decide between largely equivalent options ;)

So `pnorm` at the centre, and others as wrappers then, is that the verdict?

pmcharrison:
Well, I guess the point I wanted to make is that the three options aren't really equivalent. The route I proposed allows you to write one detailed function:

```py
def pnorm_distance(x, y, p):
    return sum((array(x) - array(y)) ** p) ** (1 / p)
```

and then two trivial functions:

```py
def euclidean_distance(x, y):
    return pnorm_distance(x, y, p=2)

def taxicab_distance(x, y):
    return pnorm_distance(x, y, p=1)
```

which I think is quite a lot better for readability and maintainability than writing three separate functions. I am enthusiastic in an academic sense about exploring code performance, but in this case I don't think I would have bothered, because fundamentally the difference between the implementations here is trivial: it's basically about whether you hardcode the 2 or not, which should make basically no difference to the computation time if the functions are otherwise implemented equivalently.

pmcharrison:
> So `pnorm` at the centre, and others as wrappers then, is that the verdict?

Yes, I think so.

MarkGotham:
like so?

amads/algorithms/norm.py
```diff
@@ -0,0 +1,208 @@
...
+from typing import Iterable
+
+import numpy as np
+
+__author__ = "Mark Gotham"
+
+
+# ------------------------------------------------------------------------------
+
+l1_names = ["sum", "manhattan", "city", "cityblock", "city block", "taxicab", "l1"]
```
pmcharrison, line 19: It's nice that you provide these options but I think more conventionally you would just give one name and then the user can work out this name by looking at the documentation!

amads/algorithms/norm.py
```diff
@@ -0,0 +1,208 @@
...
+l2_names = ["euclidean", "l2"]
+max_names = ["max", "maximum", "inf", "infinity"]
+
+
+def normalize(
+    profile: Iterable,
+    method: str = "Euclidean",
+    round_output: bool = False,
+    round_places: int = 3,
```
pmcharrison, line 28: As discussed before, I don't think we should round in these kinds of cases

amads/algorithms/norm.py
```diff
@@ -0,0 +1,208 @@
...
+    >>> manhattan_distance(profile_1, profile_2)
+    5.0
+
+    >>> manhattan_distance(profile_1, profile_2, norm_=True)
+    0.2
+
+    """
+    shared_length(profile_1, profile_2)
+    if norm_:
+        profile_1 = normalize(profile_1, "l1")
```
pmcharrison, line 132: Is there a reason we can't put this normalisation functionality in the p norm function? I suppose your current definition of `normalize` doesn't support float orders, but isn't this trivial to support?

pmcharrison (approved):
I put some minor comments but I think this is more than enough detail to be going into so I think you should just respond to what you want then merge!

PR #83, opened 2025-03-27
Add `nPVI`

HuwCheston:
This PR is the first of several that should split up the timing features introduced in https://github.com/music-computing/amads/pull/12 into multiple separate PRs. The hope is that this should be easier to keep track of discussions than a single large PR!

This PR implements `npvi.py` and adds related tests. It should also resolve all related comments from the previous PR.

### DIFF
--- a/tests/test_npvi.py
+++ b/tests/test_npvi.py
@@ -1,0 +1,84 @@
+"""Test suite for functions in amads.time.npvi"""
+
+import unittest
+
+from amads.time.npvi import normalized_pairwise_variability_index
+
+
+class NPVITest(unittest.TestCase):
+    def test_daniele_patel_2013_example_1(self):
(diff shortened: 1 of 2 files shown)


tests/test_npvi.py
```diff
@@ -0,0 +1,102 @@
+"""Test suite for functions in amads.time.npvi"""
+
+import unittest
```
pmcharrison, line 3: We use pytest, not unittest, sorry. Maybe you can ask chatgpt to rewrite in pytest?

HuwCheston: Resolved in https://github.com/music-computing/amads/pull/83/commits/f0744f7c3ee5b0380879f46265d4c5b2f33a8948

pmcharrison: Hehe that was quick!

HuwCheston: 🤖

pmcharrison (approved):

PR #85, opened 2025-03-27
Add `tempo.py`

HuwCheston:
Continuing the theme of splitting https://github.com/music-computing/amads/pull/12 into separate PRs, this adds several tempo modelling features and fixes comments made on the initial PR.

### DIFF
--- a/amads/time/tempo.py
+++ b/amads/time/tempo.py
@@ -31,7 +31,7 @@ def tempo_slope(beats: Iterable[float]) -> float:
     The tempo slope represents the overall tempo change per second in a performance.
     It is determined by the slope of a linear regression of instantaneous tempo against
     beat onset time. A negative slope indicates deceleration, while a positive slope
-    indicates acceleration.
+    indicates acceleration. The units are (quarter-note) beats-per-minute-per-second
 
     The equation is:
 
@@ -39,7 +39,7 @@ def tempo_slope(beats: Iterable[float]) -> float:
         \hat{S} = \frac{\sum\limits_{i=1}^N (x_i - \bar{x}) (y_i - \bar{y})}{\sum\limits_{i=1}^N (x_i - \bar{x})^2},
 
     where :math:`x_i` is the time of beat :math:`i` and :math:`y_i` is the tempo value in
-    quarter-note beats-per-minute.
+    (quarter-note) beats-per-minute.
 
     Parameters
     ----------
@@ -69,7 +69,7 @@ def tempo_fluctuation(beats: Iterable[float]) -> float:
     .. math::
         \text{F} = \dfrac{\sqrt{\frac{1}{N-1} \sum\limits_{i=1}^N (y_i - \bar{y})^2}}{\bar{y}},
 
-    where :math:`y_i` is the tempo value in quarter-note beats-per-minute at beat :math:`i`.
+    where :math:`y_i` is the tempo value in (quarter-note) beats-per-minute at beat :math:`i`.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


amads/time/tempo.py
```diff
@@ -1 +1,213 @@
...
+
+    The equation is:
+
+    .. math::
+        \hat{S} = \frac{\sum\limits_{i=1}^N (x_i - \bar{x}) (y_i - \bar{y})}{\sum\limits_{i=1}^N (x_i - \bar{x})^2},
+
+    where :math:`x_i` is the time of beat :math:`i` and :math:`y_i` is the tempo value in
+    quarter-note beats-per-minute.
+
+    Args:
```
pmcharrison, line 51: For consistency we should have this back in numpydoc formatting too.

amads/time/tempo.py
```diff
@@ -1 +1,213 @@
...
+
+
+# flake8: noqa: W605
+def tempo_drift(beats: Iterable[float]) -> float:
+    r"""
+    Calculates the tempo drift for a sequence of beat timestamps.
+
+    The tempo drift is the standard error of the slope of the overall tempo change in a performance,
+    such that larger values imply a greater departure from a linear tempo slope (i.e., not just
+    accelerating or decelerating).
```
pmcharrison, line 70: I don't think the standard error of the slope captures what you are describing. The SE of the slope captures how uncertain we are about the slope value; as the piece gets longer and longer, this will converge to 0. What you are talking about is captured rather by the residual variance in the tempo regression.

pmcharrison: Also, is tempo drift really an intuitive term for that? If tempo gradually slows down by 10bpm over the course of the performance, does that not count as tempo drift? I would have thought it should.

amads/time/tempo.py
```diff
@@ -1 +1,213 @@
...
+
+def beats_to_tempo(beats: np.ndarray) -> np.ndarray:
+    """Converts beat timestamps to instantaneous tempo measurements."""
+
+    # Raise error on invalid inputs
+    _validate_beats(beats)
+    return np.array(60 / np.diff(beats))
+
+
+def linregress(x: np.ndarray, y: np.ndarray) -> LinearRegressionWrapper:
```
pmcharrison, line 160: I appreciate the enthusiasm but given that we already have scipy as a dependency I think it would be safest to use `scipy.stats.lingress` here!

amads/time/tempo.py
```diff
@@ -1 +1,213 @@
...
+
+    """
+
+    return float(fit_tempo_linear_regression(beats).slope_stderr)
+
+
+# flake8: noqa: W605
+def tempo_fluctuation(beats: Iterable[float]) -> float:
+    r"""
+    Calculates the percentage fluctuation around the overall tempo of a sequence of beats.
```
pmcharrison, line 105: Looking at the code it's not a percentage

amads/time/tempo.py
HuwCheston: yes, sure. took another look in the literature and can't actually find a reference for a `tempo_drift` function. not entirely sure where i got this from as we don't use it in any of our papers either. so i think safest just to remove. the `fit_tempo_regression` function already exists and can be used to grab any of the attributes we want from the returned `scipy` object.

amads/time/tempo.py
HuwCheston: fixed in https://github.com/music-computing/amads/pull/85/commits/00d617e2964dcee007d8b7ce86e1ea7eca16663d

amads/time/tempo.py
HuwCheston: fixed in https://github.com/music-computing/amads/pull/85/commits/00d617e2964dcee007d8b7ce86e1ea7eca16663d (same for other functions too)

amads/time/tempo.py
pmcharrison: Sounds like a sensible way forward!

pmcharrison (approved):

PR #86, opened 2025-03-27
Bugfix `npvi` docstring crashing `sphinx`

HuwCheston:
Currently the docstring for `npvi` is crashing the documentation build with `sphinx`. This I believe is due to the numpy style docstring:

```
Parameters
------------
```
crashing due to adding an unexpected section in RST format. 

This fix just removes the numpy style docstring and allows the documentation build to work properly on my machine at least. ~~I've also noted that the equation isn't properly formatting, however, and would welcome any thoughts on that from @pmcharrison @MarkGotham~~

In the future, I think we should add building the documentation as part of the PR process to avoid this from happening again.

### DIFF
--- a/amads/time/npvi.py
+++ b/amads/time/npvi.py
@@ -15,41 +15,42 @@ Citation: Daniele, J. R., & Patel, A. D. (2013). An Empirical Study of Historica
 from typing import Iterable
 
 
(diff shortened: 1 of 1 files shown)


pmcharrison (approved):
I've now fixed the documentation building in PRs

PR #87, opened 2025-03-28
Add `swing.py`

HuwCheston:
This is the final PR derived from https://github.com/music-computing/amads/pull/12 that adds `swing.py` and functions related to `beat_upbeat_ratio`. Compared with the previous implementation, this works by:

- Taking in an iterables of `beats` and `upbeats`
- We leave the handling of which onsets are beats and which are upbeats up to the user as it isn't currently clear to me how we want to handle this.
- Validating both iterables (checking for not enough values, missing values, overlapping values)
- Matching consecutive beats and upbeats together:
- In cases where more than one `upbeat` can be matched with two consecutive `beats` (or no `upbeats` can be matched with them), we skip over these cases: as @pmcharrison points out in https://github.com/music-computing/amads/pull/12#discussion_r1880963470, this "multiple upbeats" situation can't really be considered swing 8ths.
- Calculating the BUR for all valid cases, adding in None for invalid cases
- Returning an array with size == `len(beats) - 1`. 

There are two utility functions which just calculate mean and SD BUR from a list of beats + upbeats. Note that these use `np.nanmean` to handle `null` values: I think this makes sense personally, as it means the user does not need to chop up a large list of beats/upbeats into multiple sublists and then call `mean_bur` on each of these. Others may disagree, however!

### DIFF
--- a/amads/time/swing.py
+++ b/amads/time/swing.py
@@ -62,25 +62,52 @@ def beat_upbeat_ratio(
     Additionally, the function can calculate the :math:`log_2` of the BUR values, where a value of 0.0
     corresponds to "triplet" swing. This can be enabled by setting `log2=True`. The values can also
     be filtered to remove outliers by setting `bounded=True`, with the default values for the boundaries
-    coming from Corcoran & Frieler (2021)
+    coming from Corcoran & Frieler (2021).
 
-    Args:
-        beats (Iterable[float]):
-            An array of beat timestamps. Should not overlap with `upbeats`.
-        upbeats (Iterable[float]):
-            An array of upbeat timestamps.
-        log2 (bool, optional):
-            If True, computes the log base 2 of BUR values, as used in [2]. Defaults to False.
-        bounded (bool, optional):
-            If True, filters out BUR values outside the specified range. Defaults to False.
-        lower_bound (float, optional):
-            Lower boundary for filtering BUR values. Defaults to 0.25 (:math:`log_2` -2).
-        upper_bound (float, optional):
-            Upper boundary for filtering BUR values. Defaults to 4.0 (:math:`log_2` 2).
+    Parameters
+    ----------
+    beats : Iterable[float]
+        An array of beat timestamps. Should not overlap with `upbeats`.
+    upbeats : Iterable[float]
+        An array of upbeat timestamps.
+    log2 : bool, optional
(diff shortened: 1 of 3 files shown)


amads/time/swing.py
```diff
@@ -1 +1,172 @@
...
+    a size of `len(beats) - 1`. If multiple `upbeats` are found between two consecutive `beats`
+    or if no `upbeat` is found, the BUR for those beats will be omitted and the corresponding value
+    will be `None`.
+
+    Additionally, the function can calculate the :math:`log_2` of the BUR values, where a value of 0.0
+    corresponds to "triplet" swing. This can be enabled by setting `log2=True`. The values can also
+    be filtered to remove outliers by setting `bounded=True`, with the default values for the boundaries
+    coming from Corcoran & Frieler (2021)
+
+    Args:
```
pmcharrison, line 67: Not sure why you switched away from numpydoc formatting? That's what the rest of amads uses

amads/time/swing.py
```diff
@@ -1 +1,172 @@
...
+            If True, filters out BUR values outside the specified range. Defaults to False.
+        lower_bound (float, optional):
+            Lower boundary for filtering BUR values. Defaults to 0.25 (:math:`log_2` -2).
+        upper_bound (float, optional):
+            Upper boundary for filtering BUR values. Defaults to 4.0 (:math:`log_2` 2).
+
+    Returns:
+        list[float]:
+            A list of the calculated BUR values.
+
```
pmcharrison, line 84: How come you got rid of the examples? They are useful to have

amads/time/swing.py
HuwCheston: Got a bit scared by the crashes I was getting with Sphinx that were related to the `----------` `numpydoc` uses to separate titles in the docstring. Any ideas?

amads/time/swing.py
HuwCheston: Moved to tests, but can move them back...

amads/time/swing.py
pmcharrison: Have pushed a reformatted version that seems to work!

amads/time/swing.py
pmcharrison: Maybe just one or two, so that the usage is illustrated? I do like the comprehensivity of the tests that you have now!

pmcharrison (approved):

PR #92, opened 2025-04-02
A start on instrument:fingering

MarkGotham:


MarkGotham:
@rbdannenberg does this look ok to you?

If so, I'll get populating for all the default pairings.

Issue #82, opened 2025-03-26
Basic, shared nomenclature.


MarkGotham:
Hi @rbdannenberg , @pmcharrison , all,

Lately I've been thinking about basic, shared nomenclature and functionality. We've discussed this in the past, I think. If we settled on something and I've forgotten then please remind me. In any case, I think I'm coming at this "fresh" (... for better or worse ;) ... )

Basically, there's a case for shared representations around a general distinction centred on the "set" and the "vector".

A set is an unordered list of elements. A standard example for music theory would be the pitch class set, though further distinction is drawn between:
- pitch set (which includes spelling, excludes duplicates),
- a pitch _class_ set (excludes spelling and duplicates) and 
- a pitch _multi_set (includes spelling and duplicates).
- a pitch class _multi_set (excludes spelling, includes duplicates).

So `class: bool` and `multi: bool`. 

These "sets" are distinct from "vectors". A "vector" is an ordered list recording the number of elements at each position. Keeping with pitch, these positions may concern pitch classes (e.g., 12-TET) or scale degrees.

We use the term "profile" as broadly synonymous with "vector". If there is a distinction it's informal and more slight, with the profile perhaps implying some kind of prototypical or referential status. Or perhaps there's no difference. A "Pitch Class Profile (PCP)" is a vector of length 12 recording the number or weighting of the pitch classes 0-12. This could encode raw count or weightings.

A "distribution" is a special case of the profile or vector, expressing the elements as a proportion of the total such that the value sum to 1.

An "indicator vector" is a simplified vector indicating only the presence or absence of elements at the given position, typically with `1` for present, and `0` for absent.

This is "shared" because almost all of the logic above applies not only to pitch, but also rhythm (in the sense of beat classes) and intervals (with some caveats). The "beat class set" encodes the number of elements at each position in some kind of temporal (e.g, metrical) cycle. (We have discussed the complexities of meter separately).

In all cases, the term "class" dispenses with equivalents. In the case of pitch class this is enharmonic spelling. In the case of interval this is octave inversion-equivalence so that interval class 1 includes 1, -1, 11, -11. As such, the "interval class set" (a recognised term in music theory) dispenses with intervals 7–11, limiting the range to 0–6. Separately from this, we can also speak of an "interval set" (again, ordered or not and "multi" or not). 

[...]

PR #93, opened 2025-04-04
Create vectors_sets.py

MarkGotham:
Following on from #82, here's a start.

I would then plan on adapting modules like [pitch/transformations.py](https://github.com/music-computing/amads/blob/main/amads/pitch/transformations.py) to use these where possible, freely adding wrapper functions so that we still have a new `pitch.transpose_set` and `transpose_vector` both calling the relevant functions from a central module under the hood.

### DIFF
--- a/amads/core/vectors_sets.py
+++ b/amads/core/vectors_sets.py
@@ -236,7 +236,7 @@ def is_indicator_vector(indicator_vector: list | tuple) -> bool:
     return False
 
 
-def weighted_to_indicator(weighted_vector: list | tuple, threshold=0.0):
+def weighted_to_indicator(weighted_vector: list | tuple, threshold=0.0) -> tuple:
     """
     Converts a weighted vector to an indicator vector.
 
@@ -254,14 +254,14 @@ def weighted_to_indicator(weighted_vector: list | tuple, threshold=0.0):
     --------
     >>> weighted_vector1 = [0.0, 0.0, 2.0, 0.0]
     >>> weighted_to_indicator(weighted_vector1)
-    [0, 0, 1, 0]
+    (0, 0, 1, 0)
 
     >>> weighted_vector2 = [0.2, 0.0, 1.5, 0.0, 0.01]
     >>> weighted_to_indicator(weighted_vector2)
-    [1, 0, 1, 0, 1]
+    (1, 0, 1, 0, 1)
 
     >>> weighted_to_indicator(weighted_vector2, threshold=0.1)
-    [1, 0, 1, 0, 0]
+    (1, 0, 1, 0, 0)
     """
     indicator_vector = []
     for weight in weighted_vector:
@@ -269,7 +269,7 @@ def weighted_to_indicator(weighted_vector: list | tuple, threshold=0.0):
             indicator_vector.append(1)
         else:
             indicator_vector.append(0)
-    return indicator_vector
+    return tuple(indicator_vector)
     # TODO consider np.where(weighted_vector > threshold, 1, 0)
 
 
--- a/amads/core/vectors_sets.py
+++ b/amads/core/vectors_sets.py
@@ -236,6 +236,43 @@ def is_indicator_vector(indicator_vector: list | tuple) -> bool:
     return False
 
 
+def weighted_to_indicator(weighted_vector: list | tuple, threshold=0.0):
+    """
+    Converts a weighted vector to an indicator vector.
+
+    Parameters
+    ----------
+    weighted_vector: A NumPy array representing the weighted vector. TODO consider use of numpy
+    threshold:  Values below this threshold will be set to 0.  This handles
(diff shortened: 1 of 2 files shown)


MarkGotham:
Further notes:
- This PR primarily serves to demonstrate a possible organisation of 
   - Core: vectors and sets
   - Pitch: the possible relation of pitch objects and collections thereof to that core.
- I'm aware of overlapping functionality with `amads.core.basics.Pitch` and elsewhere.
  - I also know you have a refactor in progress
  - Happy to dispense with the `Pitch` class here and use yours (/your new version) if that's preferred (i.e., merge the vectors and sets part without the pitch part which served as demo only)
  - Probably the best is some combination of both, e.g., `amads.core.basics.Pitch` creates a pitch object from a `keynum` (broadly, midi): see the complementary proposal here for `from_name` functionality, including multiple flats/sharps and/or octave.
- Pitch-specific functionality should probably be in `amads/pitch` rather than `basics`, right?

pmcharrison:
Looking good @MarkGotham!

MarkGotham:
Great, thanks @pmcharrison . Merging and moving discussion to an issue.

pmcharrison:
Just for transparency, I didn't actually do a code review,  I just skimmed it to see what you were doing. But it doesn't interact much with other parts of the codebase so I don't see a problem with merging.

MarkGotham:
> Just for transparency, I didn't actually do a code review, I just skimmed it to see what you were doing. But it doesn't interact much with other parts of the codebase so I don't see a problem with merging.

Ah ok. That wasn't clear to me. Looked like a sign off ;).

Re-opening to this effect with some specific suggestions wrt object type handling.

amads/core/vectors_sets.py
```diff
@@ -0,0 +1,294 @@
...
+__author__ = "Mark Gotham"
+
+
+def set_to_vector(
+    input_set: list[int] | tuple[int],
+    min_index: int | None = 0,
+    max_index: int | None = 6,
+) -> tuple:
+    """
+    Converts any "set" (list of integers) into a "vector" (count of integers organised by index).
```
pmcharrison, line 22: It's a bit confusing that "set" doesn't actually accept a Python `set` here. I would suggested replacing the type hint for `input_set` with `Iterable[int]`.

amads/core/vectors_sets.py
```diff
@@ -0,0 +1,294 @@
...
+    for item in input_set:
+        counts[item] += 1
+
+    if min_index > 0:
+        counts = counts[min_index:]
+
+    return tuple(counts)
+
+
+def vector_to_set(vector: list[int] | tuple[int]) -> tuple:
```
pmcharrison, line 61: It feels a bit strange that the set is not a set. Consider using a Python set as the output type?

amads/core/vectors_sets.py
```diff
@@ -0,0 +1,294 @@
...
+    True
+
+    """
+    return tuple(i for i in range(len(vector)) for _ in range(vector[i]))
+
+
+# Arithmetic operations: Addition/subtraction, multiplication,division
+
+
+def apply_constant(
```
pmcharrison, line 96: I wonder whether `add` might be a more intuitive name for this function

amads/core/vectors_sets.py
```diff
@@ -0,0 +1,294 @@
...
+        x * scale_factor for x in input_list
+    ]  # TODO np would be better in cases like this
+
+
+# Transformations
+
+
+def rotate(vector: list[int] | tuple[int], steps: int | None = None) -> list:
+    """
+    Rotate a list by N steps.
```
pmcharrison, line 160: Maybe 'rotate a vector'? I think it is good to be consistent about what we are calling our input object

amads/core/vectors_sets.py
```diff
@@ -0,0 +1,294 @@
...
+    """
+    if not steps:
+        steps = int(len(vector) / 2)
+
+    return vector[steps:] + vector[:steps]
+
+
+def mirror(vector: list | tuple, index_of_symmetry: int | None = None) -> list:
+    """
+    Reverse a list.
```
pmcharrison, line 198: Maybe 'reverse a vector'? I think it is good to be consistent about what we are calling our input object

amads/core/vectors_sets.py
```diff
@@ -0,0 +1,294 @@
...
+
+    Parameters
+    ----------
+    weighted_vector: A NumPy array representing the weighted vector. TODO consider use of numpy
+    threshold:  Values below this threshold will be set to 0.  This handles
+    cases where weights might be very small but not exactly zero.
+
+    Returns
+    -------
+    A NumPy array representing the indicator vector (0s and 1s).
```
pmcharrison, line 251: The types specified in the docstring are inconsistent with the type hints in the function definition

PR #96, opened 2025-04-08
Polish up Parncutt (1988) implementation

pmcharrison:


amads/harmony/root_finding/parncutt.py
```diff
@@ -0,0 +1,190 @@
...
+            "A",
+            "A#",
+            "B",
+        ]
+        plt.xticks(range(12), pitch_class_names)
+
+        # Add title
+        if title is None:
+            chord_str = ", ".join(
+                str(pc) for pc in sorted(set(p % 12 for p in self.chord))
```
MarkGotham, line 176: Set does sorting. 

```py
>>> a = [3, 1, 2]
>>> set(a)
{1, 2, 3}
```

amads/harmony/root_finding/parncutt.py
```diff
@@ -0,0 +1,190 @@
...
+            title = f"Root strengths for chord [{chord_str}]"
+        plt.title(title)
+
+        # Add grid
+        plt.grid(axis="y", linestyle="--", alpha=0.7)
+
+        plt.tight_layout()
+
+        # Check if we're in a Pytest environment
+        if os.environ.get("PYTEST_CURRENT_TEST"):
```
MarkGotham, line 187: Nice!

amads/harmony/root_finding/parncutt.py
```diff
@@ -0,0 +1,190 @@
...
+            plt.close()  # Close the figure to prevent memory leaks
+        else:
+            plt.show()
```
MarkGotham, line 190: Need a repo-wide policy on this

tests/test_parncutt_1988.py
```diff
@@ -1,8 +1,105 @@
...
+    assert (
+        analysis.get_root_strength(0) == 10 + 3 + 5
+    )  # Root + perfect fifth + major third
+    assert analysis.get_root_strength(1) == 0  # No intervals from this pc
+    assert analysis.get_root_strength(2) == 2 + 1  # Minor third + major second
+    assert analysis.get_root_strength(4) == 10  # Root from perspective of this pc
+
+
+def test_parn88_regression():
+    """Test regression against known values from Parncutt (1988): Table 4"""
```
MarkGotham, line 20: Great!

MarkGotham (approved):

MarkGotham:
Great, thanks!

I especially appreciate the "known values from Parncutt (1988): Table 4".

This is fine to merge as is IMO. Feel free to do so!

The only questions for me are over objects / AMADS plumbing. Simply another context in which we think through how all the objects inter-relate. Feel free to move this to an issue if you merge and close.

You've put this in `/harmony` and say it applies to chords. Is that always (necessarily) the case, or generally?
- If 'necessarily', should we make a `Chord` object for the purposes?
	- I'm all for keeping entry point simple, but other conversations have discouraged that.
- If 'generally', would be ever it appropriate to apply it to successive pitches in a melody?
	- If so, perhaps we take a `PitchCollection` (added recently as generic container for pitch objects).

Similarly wrt the prototypes: there's a case for including these with the key profiles. Or, perhaps more specifically, having key profiles, chord profiles and (in this case) root profiles somehow consistent. Arguably beat class profiles too (coming soon).

Plotting:
This one we should coordinate too.
- I like the `os.environ.get("PYTEST_CURRENT_TEST")` case.
- I wonder whether what the default handling should be. Some options:
	- I often use a `write_not_save` bool to toggle between show vs save.
	- Alternatively, return the `plt` and let the user handle show vs save?
- Hard to know how much of this to abstract with plotting. Much of the functionality will be shared elsewhere, but probably not quite enough to have a dedicated module that attempts to handle all cases.

That's all, thanks!

amads/harmony/root_finding/parncutt.py
pmcharrison: Good thoughts about this. I looked at what Roger has done elsewhere, and saw that he has a nice solution where he returns `plt, fig` from the function, so I've now adopted that for consistency.

amads/harmony/root_finding/parncutt.py
pmcharrison: I don't think this is true, actually, see e.g. https://stackoverflow.com/questions/3812429/is-pythons-set-stable

I think it's safest to explicitly sort here!

amads/harmony/root_finding/parncutt.py
pmcharrison: Now updated following your other prompts

pmcharrison:
Have now implemented support for Chord and PitchCollection inputs. Thanks for the suggestions! I think that's all of the review responded to now.

amads/harmony/root_finding/parncutt.py
MarkGotham: Huh! Interesting. Thanks.

MarkGotham:
Great! Happy for this to merge.

Only remaining ideas I had (not required for the merge) were:
- Structure of "pitch", "chord", "key", "harmony". Possibly the simplest and best would be to remove this "harmony" directory, move this function to one of the others (probably `pitch/` but as you see fit, see question of scope above) and delete "harmony".
- Maybe a tweak to re-name the function `get_root` (or add a new convenience wrapped function with the same name). `Terhadt` can be the default option, but leave it open and flexible to add alternatives later.

Issue #95, opened 2025-04-07
Core/pitch


MarkGotham:
Vectors and sets branch (#93) recently merged.

Bringing unresolved questions over here:

- Pitch: draft on that merge for the possible relation of pitch objects and collections thereof to that core.
- Overlapping functionality with `amads.core.basics.Pitch` and elsewhere.
- TODO chose/combine/adapt etc.
- Probably the best is some combination of both, e.g.,
    - `amads.core.basics.Pitch` creates a pitch object from a keynum (broadly, midi)
    - new merge adds `from_name` functionality, including multiple flats/sharps and/or octave.
    - Move all Pitch-specific functionality to `amads/pitch` rather than `basics`

MarkGotham:
#107

PR #97, opened 2025-04-09
Vectors sets

MarkGotham:


amads/core/vectors_sets.py
```diff
@@ -1,8 +1,13 @@
 # coding: utf-8
```
pmcharrison, line 1: How did you feel writing this with the constraint of relying on basic types? I know you originally thought it might be better to write some custom wrapper classes, e.g.

```py
class Set(set):
    def __init__(self, iterable):
        super().__init__(iterable)
        for elt in self:
            assert is_valid_element(elt)

    def is_valid_element(self, elt):
        return isinstance(elt, int)

    def to_indicator_vector(self, min_index, max_index):
        return IndicatorVector(...)

class IndicatorVector(tuple):
    def __init__(self, iterable, min_index=0):
        super().__init__(iterable)
        for elt in self:
            assert is_valid_element(elt)

    def is_valid_element(self, elt):
        return elt == 0 or elt == 1

    def to_set(self):
        return Set([...])
```

Without such classes, you are running into some problems otherwise with retaining necessary information about each object. For example, in your current implementation, you have a problem where the `min_index` for creating a vector is not retained, so we lose invertibility:

```py
>>> s = [1, 2, 3]
>>> s
[1, 2, 3]
>>> v = set_to_vector(s, min_index=1)
>>> v
(1, 1, 1, 0, 0, 0)
>>> vector_to_set(v)
(0, 1, 2)
```

If you implemented a custom `IndicatorVector` class it would be easy to keep track of `min_index` and then use it when converting back to a set.

It would then be quite natural for e.g. `PitchClassSet` to subclass `Set`.

I appreciate that you were steered earlier away from such custom classes, but personally I'm now starting to feel that, if you are going to write these helper functions, they would be most naturally encapsulated within these kinds of classes.

amads/core/vectors_sets.py
```diff
@@ -87,15 +92,17 @@ def vector_to_set(vector: list[int] | tuple[int]) -> tuple:
...
     """
-    return tuple(i for i in range(len(vector)) for _ in range(vector[i]))
+    return tuple([i for i in range(len(vector)) for _ in range(vector[i])])
 
 
-# Arithmetic operations: Addition/subtraction, multiplication,division
+# Arithmetic operations: Addition/subtraction, multiplication, division
 
 
 def apply_constant(
```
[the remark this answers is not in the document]
MarkGotham: WIP

amads/core/vectors_sets.py
MarkGotham: I see what you mean, especially where the basic classes are not there (vectors).

Creating a `Set` class seems a bit heavy handed though.

Alternatively, if this is all about the index trimming then that could be split off as separate functionality (perhaps calling an error if you trim indices which contain a non-0 entry?) and paired with reciprocal padding to fix the roundtrip you mention.

pmcharrison: I think if the plan is to provide substantial functionality for manipulating these objects, then the custom classes make a lot of sense. The nice thing about writing custom classes is that they are eminently future-proof, because you can achieve almost anything down the line through custom methods/attributes/properties. If the planned functionality is only minimal, then the motivation for having helper functions at all becomes somewhat disputable, because basic operations can be applied quite easily using list/set comprehensions, e.g.

```py
# add_constant
lst1 = [1, 2, 3]
lst2 = [x + constant for x in list]

# to indicator vector
vec = [int(x in lst1) for x in range(12)]
```

But perhaps the thing to do is experiment and just see how the code feels like to use, it can always be adjusted later I guess.

pmcharrison: I do agree that even without classes the functions enhance readability in the latter case, 

```py
vec = set_to_indicator_vector(lst1)
```
is probably more readable than 

```py
vec = [int(x in lst1) for x in range(12)]
```

The only thing with the former, though, is I find it hard to know when reading it whether the types are being dealt with properly. Are we sure that `to_indicator_vector` is expecting the kind of list input we are providing?

In contrast, 

```py
my_set = Set([1, 2, 3])
vec = my_set.to_indicator_vector()
```

feels rather safe in terms of type management. However, I also agree that there is a little ugliness to the reader in creating our own custom Set; the reader's immediate response might be 'hang on, why do I need this'?

No easy answers I guess!

amads/core/vectors_sets.py
```diff
@@ -14,7 +19,7 @@
 
 
 def set_to_vector(
```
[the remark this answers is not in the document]
pmcharrison: Is this present example not an indicator vector? If there are different types of vectors at play here, doesn't that increase the need to be explicit about which type we are talking about?

amads/core/vectors_sets.py
pmcharrison: PS just to add, it's possible to make the Set class lightweight by subclassing the preexisting set class, so you don't have to reimplement any of the core set methods, you just add whatever extra ones you need.

MarkGotham: Thanks for this comprehensive account.

Totally agree, especially that best practice is likely to emerge in the context of use. It may be that these functions are too basic to be useful anyway in which case we can retire them.

I suspect that it'll be useful to define these core operations as part of giving clarity to more complex processes with these simple mappings as a constituent step.

amads/core/vectors_sets.py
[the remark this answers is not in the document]
MarkGotham: clarified in latest pr

amads/core/vectors_sets.py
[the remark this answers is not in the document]
MarkGotham: clarified in latest pr

amads/core/vectors_sets.py
```diff
@@ -1,190 +1,249 @@
...
     return tuple(counts)
 
 
-def vector_to_set(vector: list[int] | tuple[int]) -> tuple:
+def vector_to_multiset(vector: tuple[int, ...]) -> tuple:
     """
     Converts any "vector" (count of integers organised by index)
-    to a corresponding "set" (unordered list of integers).
-    See the paper for full definitions.
+    to a corresponding "set" (unordered integers).
```
pmcharrison, line 86: Should say 'multiset'?

amads/core/vectors_sets.py
```diff
@@ -1,190 +1,249 @@
...
-    >>> roundtrip = set_to_vector(resulting_set, max_index=6)
+    >>> roundtrip = multiset_to_vector(resulting_set, max_index=6)
     >>> roundtrip
     (0, 3, 2, 1, 0, 0, 0)
 
     >>> roundtrip == test_vector
     True
 
     """
-    return tuple(i for i in range(len(vector)) for _ in range(vector[i]))
```
pmcharrison, line 90: Perhaps more concisely:

```py
return tuple(i for i, count in enumerate(test_vector) for _ in range(count))
```

amads/core/vectors_sets.py
```diff
@@ -1,190 +1,249 @@
...
+    elif isinstance(set_or_vector, list):
+        return result
+    elif isinstance(set_or_vector, set):
+        return set(result)
+    else:
+        return result
 
 
-def scalar_multiply(input_list: list, scale_factor: int = 2) -> list:
+def scalar_multiply(input_tuple: tuple, scale_factor: int = 2) -> tuple:
```
pmcharrison, line 188: There is something redundant (here and elsewhere about specifying the object type both in its name and in its type hint. I think `input` would be fine here (or `object`, or `x`)!

amads/core/vectors_sets.py
```diff
@@ -1,190 +1,249 @@
...
     This serves equivalently for
     "phase shifting" of rhythm and
     "transposition" of pitch.
 
     Parameters
     ----------
-    vector: Any list of any elements. We expect to work with a list integers representing a vector.
+    vector : Union[tuple[int, ...], list[int]]
+        Any tuple or list of any elements.
+        We expect to work with a list integers representing a vector.
```
pmcharrison, line 226: typo

pmcharrison (approved):

amads/pitch/__init__.py
```diff
@@ -34,7 +34,7 @@ def __init__(
             <Base name><Sharps_flats><Octave>
             (e.g., "C4", "A#", "Bbb3").
             The octave may or may not be specified and any number of sharps/flats is permitted.
-            If the octave is not be specified, the MIDI number will be assiged in octave 4,
+            If the octave is not specified, the MIDI number will be assiged in octave 4,
```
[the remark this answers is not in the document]
MarkGotham: Not working on the pitch side during #95 , #104

PR #100, opened 2025-04-15
Help with tests

pmcharrison:


pmcharrison:
@rbdannenberg if we remember later today we can perhaps discuss how to standardise the plotting interface.

PR #106, opened 2025-04-16
Plotfix - more consistent interface for plotting

rbdannenberg:
As we discussed, plot methods return fig and show plots by default, with an optional show=False parameter to simply create and return the Figure (but not (plt, fig)) for further refinement. I found multiple cases of modules that plot directly -- these should be consolidated to return a Distribution, which already has a general (enough) API to make all the plots we use, but reorganizing and extending Distribution with default axis labels, etc., is a larger effort for future work.


pmcharrison:
Closed in favour of #107

pmcharrison:
I got confused, this is not the right branch to close! Reopening.

pmcharrison (approved):
The changes look good to me!

PR #107, opened 2025-04-17
Modified amads/core/pitch.py based on amads/pitch/pitch.py

rbdannenberg:
This is a *draft* for changes to pitch.py -- it is not tested, and I don't think it's even compatible with basics.py yet, but here for comments on the interface, especially Pitch.__init__(). **Do not merge the initial pull request.**


MarkGotham:
Thanks @rbdannenberg . I'll have a go at getting this closer.

MarkGotham:
Wrt enharmonics, how about something simple like:

```py
def to_simplest_enharmonic(self, sharp_not_flat: bool = True) -> "Pitch":
    """
    Create a new Pitch object with the simplest enharmonic representation of the current pitch.
    I.e., if there exists an enharmonic-equivalent pitch with no-alterations, then use that.
    If an alteration is needed, then use sharps or flats depending on the `sharp_not_flat` attribute.

    Parameters
    ----------
    sharp_not_flat
        This is only relevant if that the pitch needs an alteration, otherwise it is unused.
        If True, then use the sharp form (e.g., 'D#').
        If False, use the flat form (e.g., 'Eb').

    Examples
    --------

    >>> es = Pitch("Eb4")
    >>> es
    Pitch(name='Eb', keynum=63)

    >>> es.alt
    -1

    >>> es.to_simplest_enharmonic()
    Pitch(name='D#', keynum=63)

    >>> es.to_simplest_enharmonic(sharp_not_flat=False)
    Pitch(name='D#', keynum=63)

    Returns
    -------
    Pitch
        A Pitch object representing the upper enharmonic equivalent.
    """
    if self.alt in [None, 0]:
        return self

    if self.pitch_class in [0, 2, 4, 5, 7, 9, 11]:  # C, D, E, F, G, A, B
        return Pitch(self.keynum)
    else:  # unaltered in 1, 3, 6, 8, 10
        if sharp_not_flat:  # sharp
            return Pitch(self.keynum, alt=1)
        else:  # flat
            return Pitch(self.keynum, alt=-1)


```

MarkGotham:
And with the test, it's mostly just `keynum` vs `key_num`.

I don't feel strongly either way, but of course it should be consistent across the repo.

PR #111, opened 2025-04-21
Rename key_num to num everywhere.

rbdannenberg:
Global rename. As a bonus, I like the slightly shorter/cleaner representation of Pitch: Pitch(name='C4', num=60)


rbdannenberg:
We revisited and decided on midi_num, which is more verbose but suggests some semantics. "num" to name an "int" does not seem such a good idea in retrospect.

PR #112, opened 2025-04-22
Add example notebook for jazz recordings using `JTD`

HuwCheston:
This PR (mainly) adds a notebook giving some examples for using features in AMADS for a simple classification task (performer identification using the Jazz Trio Database).

It also does the following:
- Small bug fix inside `amads.time.swing`
- Adds optional dependencies for `mirdata` and `sklearn` inside `pyproject.toml`.

### DIFF
--- a/amads/time/swing.py
+++ b/amads/time/swing.py
@@ -131,7 +131,16 @@ def beat_upbeat_ratio(
     # Apply our filtering if required
     # Filter before log_2 transform to make things simpler
     if bounded:
-        burs = [i if lower_bound < i < upper_bound else None for i in burs]
+        newburs = []
+        for b in burs:
+            if isinstance(b, float):
+                if lower_bound < b < upper_bound:
+                    newburs.append(b)
+                else:
+                    newburs.append(None)
+            else:
+                newburs.append(None)
+        burs = newburs
     # Express as base-2 log if required
     if log2:
         burs = [log2_(b) if b is not None else None for b in burs]
--- a/examples/classify_jazz_recordings.py
+++ b/examples/classify_jazz_recordings.py
@@ -1,0 +1,278 @@
+"""
+Jazz classification
+===================
+
+This example demonstrates how to perform a simple classification of jazz recordings
+from the Jazz Trio Database using a range of features extracted with AMADS.
+
+We'll train a simple binary classifier to identify whether a recording is by Bill Evans
+or Oscar Peterson using a combination of different melodic and rhythmic features
+"""
+
+# %%
+# First, we'll import the required modules.
+
+import random
+
+import matplotlib.pyplot as plt
+import mirdata
+import numpy as np
+from pretty_midi import PrettyMIDI
+from sklearn.impute import SimpleImputer
+from sklearn.linear_model import LogisticRegression
+from sklearn.metrics import ConfusionMatrixDisplay, confusion_matrix
+from sklearn.model_selection import train_test_split
+from sklearn.preprocessing import StandardScaler
+
+from amads.algorithms import lz77_complexity
+from amads.time.swing import beat_upbeat_ratio
+from amads.time.tempo import tempo_mean, tempo_slope
+
+# %%
+# Set some constants
+N_TRACKS = 50  # we'll only analyse this many tracks per performer
+SEED = 42
+
+np.random.seed(SEED)
+random.seed(SEED)
+
+
+# %%
+# Download the Jazz Trio Database (MIDI only) using mirdata
+# Note that this will only download annotations (beats, MIDI)
+# and will not download the audio files directly
+
+jtd = mirdata.initialize("jtd")
+jtd.download()
+
+# %%
+# Now, we load both the tracks (single instrument data, i.e. just piano)
+# and the multitracks (data for all insturments, e.g. beat tracker results)
+
+jtd_tracks = jtd.load_tracks()
+jtd_multitracks = jtd.load_multitracks()
+
+# %%
+# Print out information for a single random track
+print(jtd.choice_track())
+
+# %%
+# We're going to train a simple binary classifier (support vector machine)
+# to identify whether a recording is by Bill Evans or Oscar Peterson.
+# First things first, we need to get all of our Evans and Peterson tracks
+
+evansb_tracks = []
+petersono_tracks = []
+
+for mtrack_id in jtd.mtrack_ids:
+    associated_multitrack = jtd_multitracks[mtrack_id]
+    # We can grab the pianist data easily from the multitrack
+    associated_track = associated_multitrack.piano
+    # Store the track inside the list if required
+    if associated_track.musician == "Bill Evans":
+        evansb_tracks.append(associated_multitrack)
+    elif associated_track.musician == "Oscar Peterson":
+        petersono_tracks.append(associated_multitrack)
+    else:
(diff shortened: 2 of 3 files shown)


examples/classify_jazz_recordings.py
```diff
@@ -0,0 +1,278 @@
...
+    else:
+        continue
+
+evansb_tracks = random.sample(evansb_tracks, N_TRACKS)
+petersono_tracks = random.sample(petersono_tracks, N_TRACKS)
+
+# %%
+
+
+def skyline_prettymidi(pm: PrettyMIDI, threshold: float = 0.1):
```
pmcharrison, line 85: I think it is a bit problematic to use prettymidi and a custom skyline algorithm here, I think we should use our own package's MIDI importer and the skyline algorithm implemented already in AMADS. Otherwise this becomes something of an anti-example for the package, no?

I know that MIDI import has been slow, but hopefully it will be fixed by #116? You could let the example run slowly for now and then we can speed it up once #116 is merged I guess. Alternatively I suppose you could document the problems about slowness and provide the code that should be usable once the fixes are made.

In general, if it's impossible to run sufficiently performant code using the package's representations, this needs to be recognised and acted on.

examples/classify_jazz_recordings.py
```diff
@@ -0,0 +1,278 @@
...
+    """Applies skyline to a pretty MIDI object, following format established in amads.melody.skyline but quicker"""
+    skyline = []
```
pmcharrison, line 87: The skyline algorithm presented here is actually different from the one you used in your paper. Maybe that's worth commenting on? Would you consider updating the implementation to incorporate the different methods you have considered?

examples/classify_jazz_recordings.py
```diff
@@ -0,0 +1,278 @@
+"""
```
pmcharrison, line 1: Currently examples are only compiled into the documentation if their filename begins with `plot_`. I know, it's silly.... but best addressing so that your example builds.

pyproject.toml
```diff
@@ -48,6 +48,10 @@ melsim = [
     "pandas",
     "rpy2"
 ]
+examples = [
```
pmcharrison, line 51: For me, mirdata only installs on Python < 3.12, due to a problem with jams using removed importlib.... I'm not sure what we should do about that? Only run the CI tests on Python < 3.12?

pmcharrison (commented):
Thanks, the contribution is looking nice!

pyproject.toml
pmcharrison: Ah, it is probably because they have not made a release to PyPi yet. A good solution for now I think would be to go to pyproject.toml and write something like

```toml
[tool.hatch.metadata] 
allow-direct-references = true

[project]

dependencies = [
    # Temporary fix while we wait for mirdata > 0.3.9 to be released 
    "mirdata @ git+ssh://git@github.com/mir-dataset-loaders/mirdata@82b9710",
]
```

pmcharrison:
Just checking if you saw my other comments @HuwCheston?

HuwCheston:
Hey, thanks for the reminder @pmcharrison , will try and take a look when time permits!

amads/time/swing.py
```diff
@@ -131,13 +131,29 @@ def beat_upbeat_ratio(
...
     return burs
 
 
+def filter_burs(
+    burs: list[float], lower_bound: float, upper_bound: float
+) -> list[float]:
+    """Filters a list of BURS between lower_bound < BUR < upper_bound."""
+    newburs = []
+    for b in burs:
+        if isinstance(b, float):
```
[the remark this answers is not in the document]
HuwCheston: good call, was thinking we might need to account for cases where `b` is not a float, but that shouldn't happen when calling `beat_upbeat_ratio`

pyproject.toml
HuwCheston: done in https://github.com/music-computing/amads/pull/112/commits/f817169cb4d217e081264571723307ee39898655

examples/classify_jazz_recordings.py
HuwCheston: done in https://github.com/music-computing/amads/pull/112/commits/f817169cb4d217e081264571723307ee39898655, went for `plot_jazz_recordings.py`

amads/time/swing.py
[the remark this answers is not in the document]
HuwCheston: done in https://github.com/music-computing/amads/pull/112/commits/f817169cb4d217e081264571723307ee39898655

examples/classify_jazz_recordings.py
```diff
@@ -0,0 +1,278 @@
...
+# Download the Jazz Trio Database (MIDI only) using mirdata
+# Note that this will only download annotations (beats, MIDI)
+# and will not download the audio files directly
+
+jtd = mirdata.initialize("jtd")
+jtd.download()
+
+# %%
+# Now, we load both the tracks (single instrument data, i.e. just piano)
+# and the multitracks (data for all insturments, e.g. beat tracker results)
```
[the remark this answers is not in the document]
HuwCheston: done in https://github.com/music-computing/amads/pull/112/commits/f817169cb4d217e081264571723307ee39898655

amads/time/swing.py
```diff
@@ -131,13 +131,29 @@ def beat_upbeat_ratio(
...
     if bounded:
-        burs = [i if lower_bound < i < upper_bound else None for i in burs]
+        burs = filter_burs(burs, lower_bound, upper_bound)
     # Express as base-2 log if required
     if log2:
         burs = [log2_(b) if b is not None else None for b in burs]
     return burs
 
 
+def filter_burs(
```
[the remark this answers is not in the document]
HuwCheston: done in https://github.com/music-computing/amads/pull/112/commits/f817169cb4d217e081264571723307ee39898655

examples/classify_jazz_recordings.py
HuwCheston: ok, makes sense. for now, i'll revert to using `partitura_midi_import` here, and then when #116 gets merged, we can incorporate that here. would rather not step on @rbdannenberg 's toes with that PR.

examples/classify_jazz_recordings.py
HuwCheston: fair enough, tbh I think the current skyline algorithm in AMADS is fine for here as a demonstration of the package functionality. I can make another PR for the modified version when I have time. But yeah, I can make a comment if you think it'll enhance readability

HuwCheston:
@pmcharrison , now made requested changes. Things left to discuss:
- MIDI import is currently very slow, but merging #116 should improve things
- reverted to using the default `skyline` in AMADS as I think this example should be a demo of the package functionality. I can make a separate PR for the modified skyline when I get some time.

pmcharrison:
Thanks Huw! The code looks good but it's too slow to merge currently. What would you prefer -- to use your previous MIDI import but with the desired future code commented out, or to delay the PR until Roger's PR is merged?

HuwCheston:
sure, no worries. probably preferable to delay merging this PR until #116 is merged, as otherwise we'll need to make another PR to replace my workaround with the proper `amads` code.

pmcharrison:
Sounds good. I'll make a note to revisit in a bit! Thanks very much for your work

PR #113, opened 2025-04-23
Implement Score.get_notes

dmwhyatt:
In consultation with @pmcharrison this morning - we have been considering an easy way to stitch together code from other repositories that does not use the `Score` object. We thought a decorator that automatically extracts the relevant attributes from the `Score` would greatly simplify the process of enforcing the `Score` as the top level input, and reduce the amount of boilerplate needed to contribute existing work to AMADS. 

Hopefully, this would serve as a simple interface that makes it easier to contribute to AMADS.

Included is a demo to provide an example of how this decorator could be useful.


Dav8Circle:
@rbdannenberg I noticed that the test for `core.utils` was still using Unittest and had a typo in it, so I've updated it in line with the rest of the project, and renamed it to match!

rbdannenberg:
I have to say I don't think this is a good direction. I was going to see what it would take to express this directly, thinking it would be a big rewrite, but in fact, it's just something like:
```
    notes = score.find_all(Note)
    pitches = [note.num for note in notes]
```
and the obvious extension to other attributes, e.g. `durations = [note.dur for note in notes]`

The reason I don't like this preprocess idea is this adds another abstraction layer, so I'd argue it doesn't really help contributors and makes the basic representation more obscure with fewer examples to follow. It also is extracting attributes that many algorithms will not use and may be the wrong attributes in the first place.

This seems like another step down the path of splitting the representation into the "object-oriented camp" and the "parallel arrays of numeric data (matlab, nympy)" camp. We talked about this before and my first assumption was that's how it would work: e.g., we'd have structured data sources (scores) with one representation and analysis algorithms based on another representation, e.g. arrays, and we'd have some translation layer, perhaps doing conversions and caching the conversions, but as we implemented some of the Matlab MIDI Toolbox algorithms, it seemed that this was overkill, and iterating over notes that keep attributes  clearly named and together (pitch, onset, duration, etc.) made things more readable and organized than having arrays of numbers.

Another problem for me, and this is more just about personal programming style, is I don't think decorators are a good idea. This is basically Python's approach to macros or Lisp DEFMACRO, which IMO have a place if you want to change the language, i.e., you don't want to use Python directly, but neither do you want to define a new language, so you just redefine parts of Python. If you really need it and it has high payoff, fine, but the drawback is that even people who know Python can't read or write code (or contribute) until they learn how you've changed the semantics, and they have to keep reminding themselves because the whole point of decorators is to make something that's mostly Python not always behave like "real" Python.

If it seems better, I could see extending find_all() to return properties from objects as an option, e.g.,
```
pitches, durations = score.find_all(Note, 'num', 'dur')
```
This would extend to, e.g. retrieving a list of measure durations:
```
measure_durs = score.find_all(Measure, 'dur')
```
or find which instruments are used in the score:
```
instruments = set(score.find_all(Part, 'instrument'))  # May contain None
```
I think this would make it clear exactly where the numbers come from. E.g., there's unfortunately a difference between accessing notes in time order or in order of Part and Staff, to trying to build a one-size-fits-all extractor and hiding the details rather than offering building blocks will likely result in a proliferation of "simplifications" that are just confusing. (I know, starting with scores separated into parts and staves and potentially containing tied notes is already confusing and presents a challenge, but there are operations to remove these details and produce a simpler more "canonical" note list -- and we can provide better examples for what to do and what to expect.)

pmcharrison:
The goal was to make it easier to write functions that could take either a Score or the more basic representations as the input. We thought it would be helpful to be able to pass the basic representations directly to save repeating the score preprocessing step when applying multiple functions in sequence. In such a situation, the decorator saves a few more lines:

```py
if score is not None:
    assert pitches is None
    assert durations is None
    score = score.flatten()
    notes = score.find_all(Note)
    pitches = [note.num for note in notes]
    durations = [note.dur for note in notes]
```

I quite like your `score.find_all(Note, 'num', 'dur')`. We lose the flattening and the assertions, I suppose it's a trade off...?

rbdannenberg:
I think there's little advantage to writing algorithms in terms of lists of numbers as opposed to lists of Notes, so a step to extract numbers from Note objects is adding overhead to change representations, but if you are trying to reuse existing implementations using lists of numbers, I can see wanting to keep the list-of-number parameters as an option (although it seems unlikely it's going to make much difference). Assuming that's a goal, throwing in score as a parameter and writing
```
if score is not None:
    pitches, durations = score.find_all(Note, 'num', 'dur')
```
seems like a good compromise -- I think it's pretty compact, readable, and intuitive. It's not really boilerplate if it's 2 lines of code.

rbdannenberg:
Yes, stripping ties is still linear, but requiring multiple redundant sweeps over the score seems like a bad design. Some preliminary ideas:

- we should be explicit about expected score properties (tied notes?, flattened?)
- we could store the constraints that the score satisfies to make checking fast
- algorithms could
  - test that the score satisfies whatever preconditions it has, or
  - we could write something like `score = Score.must_be_flattened(score)` where the conversion only takes place if the constraint is not already met
- but it's more efficient if you derive the Score you need after importing the score (or maybe even as an import option), so maybe functions like `Score.must_be_flattened` should print a warning so you can see if you are performing the same conversion in multiple analysis functions instead of in one place before you call them.

pmcharrison:
Thank you, this is useful!

Following your suggestion above, we could check a `has_ties` attribute at the beginning of the function:

```py
def find_leaps(score: Score) -> list[tuple[int, int]]:
    """Find all melodic leaps (intervals larger than a major 3rd)."""
    assert not score.has_ties
    notes = score.find_all(Note)
    leaps = []
    for note_1, note_2 in zip(notes, notes[1:]):
        interval = abs(note_2.num - note_1.num)
        if interval > 4:
            leaps.append((note_2.onset, interval))
    return leaps
```

This works pretty well I think. The only thing is that it would be quite easy for a contributor to forget to check the `has_ties` argument. Maybe an alternative could be to add a `get_notes` method that forces the user to specify whether they want ties and/or sorting by time:

```py
def find_leaps(score: Score) -> list[tuple[int, int]]:
    """Find all melodic leaps (intervals larger than a major 3rd)."""
    notes = score.get_notes(strip_ties=True, sort=True)
    leaps = []
    for note_1, note_2 in zip(notes, notes[1:]):
    ...
```

We could cache the `get_notes` function so that repeated computation is not expensive.

I note that we already have a method called `get_sorted_notes` which (according to the docstring) already strips ties. Maybe this `get_notes` method could be the improved version of that?

rbdannenberg:
yes, I'm all for reviewing methods and method names for accessing scores and dealing with different "standard" score representations (e.g., flattened, tied notes), ideally based on a survey of what algorithms are actually using or requiring. Maybe for analysis the only thing we do is look at scores as a "piano roll," but I would expect there's also a need at least for analyzing instrumental parts separately.

pmcharrison:
Sounds good, so perhaps we can suggest that @Dav8Circle drops this new decorator for now, and trials instead a `.get_notes` method like the one above? 

@Dav8Circle, for porting your existing functions over, you might want to have a second look at functions where you create two separate lists (e.g. one of pitches, one of onsets), and see if it is more succinct to write them instead as one list of notes. For example, `find_leaps` is currently written like this:

```py
    leaps = []
    for i in range(len(pitches)):
        interval = abs(pitches[i] - pitches[i - 1])
        if interval > 4:  # larger than major 3rd
            leaps.append((onset_times[i], interval))
    return leaps
```

but you could consider instead writing it like this:

```py
    leaps = []
    for note_1, note_2 in zip(notes, notes[1:]):
        interval = abs(note_2.num - note_1.num)
        if interval > 4:
            leaps.append((note_2.onset, interval))
    return leaps
```

That way you wouldn't need to write extra lines to create the `pitches` and `onsets` lists.

Note incidentally that there is a a bug in your original implementation: when `i == 0`, `pitches[i - 1]` wraps around to the end of the sequence, meaning that you compute an interval between the first note and the last note.

PR #115, opened 2025-04-28
Create grid.py

MarkGotham:


amads/time/meter/grid.py
```diff
@@ -0,0 +1,181 @@
...
+    >>> starts_to_measure_relative_counter(test_list)
+    Counter({0.0: 5, 0.5: 2, 0.75: 1, 0.33000000000000007: 1, 0.6669999999999998: 1})
+    """
+    for item in starts:
+        if not isinstance(item, (int, float)):
+            raise TypeError(
+                f"All items in `starts` must be numeric (int or float). Found: {type(item)}"
+            )
+
+    return Counter([(x - int(x)) for x in starts])
```
pmcharrison, line 36: This seems like a case where you might want to round to a set number of digits?

amads/time/meter/grid.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def metrical_gcm(
```
pmcharrison, line 39: Does this function assume that all the bars in the passage are the same length? Maybe worth stating explicitly

amads/time/meter/grid.py
```diff
@@ -0,0 +1,181 @@
...
+    starts: Union[Iterable, Counter],
+    bins: int = 12,
+    distance_threshold: float = 1 / 24,
+    proportion_threshold: float = 0.999,
+):
+    """
+    To create a grid featuring every metrical position used in a source,
+    we need to find the greatest common multiple (GCM).
```
pmcharrison, line 47: I see greatest common multiple, lowest common multiple, and greatest common divisor being used seemingly interchangeably, is this a mistake?

amads/time/meter/grid.py
```diff
@@ -0,0 +1,181 @@
...
+    cumulative_count = 0
+
+    for x in counter_starts:
+        test_case = x * bins
+        diff = abs(round(test_case, 1) - test_case)
+
+        logging.debug(f"Start position {x}. Testing against {bins} bins. Gap of {diff}")
+        if diff > distance_threshold:
+            logging.debug(f" ... is outside the threshold of {distance_threshold} ...")
+            for multiplier in [2, 3, 4, 6, 8, 12, 32, 96]:
```
pmcharrison, line 151: This algorithm doesn't look so efficient, and it only tries a finite number of options?

I wonder about instead something like this:

- We have a list of metrical positions, e.g. `[a, b, c, d]`
- We wish to find the length of the tatum. The tatum can be defined as GCD(a, b, c, d) where GCD is applied with a specified tolerance.
- An example implementation of `GCD(a, b)` with tolerance can be found [here](https://stackoverflow.com/a/45325587)
- GCD is associative so `GCD(a, b, c, d) = GCD(a, GCD(b, GCD(c, d)))` and can therefore be computed as `reduce([a, b, c, d], GCD)`

pmcharrison (commented):
Seems like a useful function to have! I suppose I might have called it `find_tatum`.

amads/time/meter/grid.py
MarkGotham: Hi @pmcharrison ! 

Thanks for the suggestion. I see the logic, and am all for efficiency where there's no cost, however, in practice:
- these multipliers (2x, 3x ...) are almost always what we're after
- setting the order explicitly controls the preferred return sequence (e.g., if 2x is good enough then ignore the rest)
- _not_ using them seems to leave the system open to accumulating floating point errors, which can easily lead to suggesting the use of 93 divisions per measure instead of 96. This may be mathematically correct but typically misses the musical point.

All of this serves to reverse engineer information lost in recording events positions in floats. An alternative would be to go after the source issue, but that's a big project, given how many datasets there are with events at position `X.83333...`

Another work-around is to scale (by 1000, say) and re-run on integers, but this is basically a worst-of-both-worlds. It would be better to do that scaling by the 2x and 3x multiples, which leads back to initial design above.

So, I'm inclined to stick with the initial design, perhaps formalising the 2x, 3x multiples a bit, perhaps as follows:

```py
def iterate_powers(limit, bases=(2, 3)):
    """
    Iterates over powers of multiple bases up to a given limit.

    Parameters
    -----------
    limit: The upper bound (inclusive) for the powers.
    bases: A tuple of integers representing the bases for calculating powers.
             Defaults to (2, 3).
    """
    seen = set()  # Track powers we've already yielded

    for base in bases:
        power = 0
        while True:
            result = base ** power
            if result > limit:
                break
            if result not in seen:
                yield result
                seen.add(result)
            power += 1
```

amads/time/meter/grid.py
MarkGotham: True. Mea culpa. Fixed.

amads/time/meter/grid.py
MarkGotham: True. Clarified.

amads/time/meter/grid.py
MarkGotham: Ok, with caveat as below.

amads/time/meter/grid.py
MarkGotham: You're right about this little power function. That's not a good idea, I retract it. Any such list should simply be explicit as currently, though probably promoted to an argument with user-defined option (if they want base 5, say) and with a default (broadly as previously).

pmcharrison: Hi @MarkGotham, just making a brief response now before I have time look through -- your idea is actually a very good one, to preprocess all measure-relative positions to approximate them as fractions. I can just suggest a slight modification, which is that we automate that process from first principles rather than doing it via a lookup table. Here is a function I wrote for that in R a while back (it's similar to the GCD algorithm):

```py
# See DOI: 10.1080/17459737.2015.1033024
# @param x Float to approximate
# @param d Tolerance ratio
# @return A two-element list corresponding to the numerator and denominator of the fraction
fraction <- function(x, d, verbose = FALSE) {
  x_min <- (1 - d) * x
  x_max <- (1 + d) * x
  a_l <- floor(x)
  b_l <- 1
  a_r <- floor(x) + 1
  b_r <- 1
  a <- round(x)
  b <- 1
  while(a / b < x_min || x_max < a / b) {
    x_0 <- 2 * x - a / b
    if (x < a / b) {
      a_r <- a
      b_r <- b
      k <- floor((x_0 * b_l - a_l) / (a_r - x_0 * b_r))
      a_l <- a_l + k * a_r
      b_l <- b_l + k * b_r
    } else {
      a_l <- a
      b_l <- b
      k <- floor((a_r - x_0 * b_r) / (x_0 * b_l - a_l))
      a_r <- a_r + k * a_l
      b_r <- b_r + k * b_l
    }
    a <- a_l + a_r
    b <- b_l + b_r
    if (verbose) message("a = ", a, ", b = ", b)
  }
  c(a, b)
}
```

So, you could perform all the fractional approximations first, then it becomes trivial to find the common divisors for those fractions.

pmcharrison: This would avoid the compounding error problem of my previous suggestion.

pmcharrison: This is very interesting @MarkGotham, thank you! The docstring examples are helpful. It sounds like an interesting problem to pursue further in the future but for now I agree that it makes sense to export your solution and perhaps add some notes to the docstring about the issues that we have discussed, specifically limitations in terms of the candidates that are considered, the efficiency of the computation, and the inadequacy of seemingly obvious alternate implementations using the Euclidean algorithm. Your algorithm and its test cases can then be a useful benchmark for future work trying to improve on it.

MarkGotham: Great, thanks @pmcharrison . (And you're right about the name change btw).

For any mathematicians following along the problem is "Euclid but with a preference for some denominators over others".

All done now. Ready to merge, as you say, on a provisional basis.

pmcharrison: Seems a sensible way of going forward. I guess one final check to make on my end is -- is it possible that the problems you observe only happen when working with small numbers of decimal places? Is it possible that when working with 'real' symbolic scores the precision is so high that it's easy to identify the correct fractional approximation?

pmcharrison: because in most cases, the score itself does contain the 'real' fraction, no? a radical solution would be for us to actually encode this fraction (e.g. https://docs.python.org/3/library/fractions.html)

pmcharrison: but I suppose you are not thinking of such cases, because such cases already contain a time signature? what is the use case you have in mind exactly?

MarkGotham: Thanks for this @pmcharrison. 

I agree that:
- This is a non-issue in the case of real score with fractions (no floats and no rounding).
- That distinction warrants clarification. I’ve added comments to the code under review (also moving most of the discussion to the top of the module). As part of this, I’m also keen to hear from your experience. If you’re engaging scores and fractions directly, do you find you need the tolerance? Is Euclid not enough? I.e., what was your use case?
- Also agree that it would be good to test the relationship between rounding tolerance thoroughly.

The use case for mine is any source that represents musical time in floats or any pipeline that includes such a step. Examples include:
- Secondary representations like the csvs in datasets like the BPSD case I’ve demonstrated. (BPSD actually also include musicxml and more alongside the csvs, though many don’t).
- Any use of MIDI or similar representations based on a more frame-wise representation of time. Likewise, by extension any transcription from audio and other sources that are frame based / more continuous.
- Any processing via code libraries that convert fractions to floats.

Altogether, many contexts rely on at least one of the above, making this is a more than peripheral set of use cases.

pmcharrison:
Hi @MarkGotham, interesting, thank you!

> any source that represents musical time in floats or any pipeline that includes such a step

In my mind, floats normally have at least 8 digits of precision (though the default in Python is 16). Shouldn’t 8 be enough? In the CSV datasets you are talking about, how much precision do they have?

I would have thought it’s hard to apply this algorithm at all to performance-timed MIDI, as that typically doesn’t have bar numbers, and I would have imagined that systematic microtiming effects would make this kind of arithmetic approach rather brittle. Maybe those kinds of cases would need probabilistic modelling instead?


From: Mark Gotham ***@***.***>
Date: Tuesday, 13 May 2025 at 20:59
To: music-computing/amads ***@***.***>
Cc: Peter Harrison ***@***.***>, Mention ***@***.***>
Subject: Re: [music-computing/amads] Create grid.py (PR #115)

@MarkGotham commented on this pull request.

________________________________

In amads/time/meter/grid.py<https://github.com/music-computing/amads/pull/115#discussion_r2087537496>:

> +                )

+        counter_starts = starts

+    else:  # Convert to Counter (also includes type checks)

+        counter_starts = starts_to_measure_relative_counter(starts)

+

+    total = sum(counter_starts.values())

+    cumulative_count = 0

+

+    for x in counter_starts:

+        test_case = x * bins

+        diff = abs(round(test_case, 1) - test_case)

+

+        logging.debug(f"Start position {x}. Testing against {bins} bins. Gap of {diff}")

+        if diff > distance_threshold:

+            logging.debug(f" ... is outside the threshold of {distance_threshold} ...")

+            for multiplier in [2, 3, 4, 6, 8, 12, 32, 96]:

Thanks for this @pmcharrison<https://github.com/pmcharrison>.

I agree that:

  *   This is a non-issue in the case of real score with fractions (no floats and no rounding).
  *   That distinction warrants clarification. I’ve added comments to the code under review (also moving most of the discussion to the top of the module). As part of this, I’m also keen to hear from your experience. If you’re engaging scores and fractions directly, do you find you need the tolerance? Is Euclid not enough? I.e., what was your use case?
  *   Also agree that it would be good to test the relationship between rounding tolerance thoroughly.

The use case for mine is any source that represents musical time in floats or any pipeline that includes such a step. Examples include:

  *   Secondary representations like the csvs in datasets like the BPSD case I’ve demonstrated. (BPSD actually also include musicxml and more alongside the csvs, though many don’t).
  *   Any use of MIDI or similar representations based on a more frame-wise representation of time. Likewise, by extension any transcription from audio and other sources that are frame based / more continuous.
  *   Any processing via code libraries that convert fractions to floats.

Altogether, many contexts rely on at least one of the above, making this is a more than peripheral set of use cases.

pmcharrison:
Sorry, email footer removed!

Yes, I've seen your case studies, and they are very helpful for understanding the behaviour of the algorithm, but they are all expressed to 3 decimal places which I thought was unnecessarily imprecise. But is the point that three decimal places corresponds approximately to MIDI resolution?

MarkGotham:
Thanks @pmcharrison . 3dp is out there! These values are from a reputable dataset.

Overall, I see two questions:
1. Is there a use case? I think I've shown that there is.
2. Is there an existing, better algorithm? Not that I have found. If it turns out there is one then I'll happily stand aside and make way for it!

pmcharrison:
Good to know! I think we should move forward then.

amads/time/meter/grid.py
```diff
@@ -0,0 +1,181 @@
...
+__author__ = "Mark Gotham"
+
+
+# ------------------------------------------------------------------------------
+
+import logging
+from collections import Counter
+from typing import Iterable, Union
+
+logging.basicConfig(
```
pmcharrison: Should we remove this logging configuration now?

pmcharrison (approved):

PR #116, opened 2025-04-29
Fastmidiin

rbdannenberg:
This branch adds code to import using Music 21. Partitura is very slow due to code that enumerates subclasses (reported to Partitura project, so I'm hoping they will find a fix.) The new code passes tests, which now run in 6.3s. Note that Music21 does not parse all musicXML examples correctly or at all. Partitura seems more robust, and doesn't parse all examples correctly either. @pmcharrison please see change to sliding_window -- before changes, adjacent but non-overlapping notes would be included in window. And now, it looks like the first window (entirely before the starting point) will always be empty. Maybe the correct thing is to drop the first window and start with the window overlapping the region of interest by the hopsize.


pmcharrison:
Sorry for slow reply @rbdannenberg -- am swamped at the moment but hope to look soon.

pmcharrison:
Hi @rbdannenberg, can you explain the rationale for for changing the note inclusion policy for `sliding_window`? The new policy seems a bit strange to me. For example, if I were to have a Score comprising solely quarter notes and applied a sliding window with size = 1.0 and step = 1.0, producing windows of [-1, 0], [0, 1], [1, 2], [2, 3], then all the sliding windows would come up empty.

amads/io/scoreread.py
```diff
@@ -0,0 +1,119 @@
...
+    midi_import, xml_import: functions for importing MIDI and XML files
+    """
+    if preferred_reader == "music21":
+        try:
+            print("In scoreread: importing for music21")
+            from amads.io.m21_midi_import import music21_midi_import
+            from amads.io.m21_xml_import import music21_xml_import
+
+            return music21_midi_import, music21_xml_import
+        except ImportError:
```
pmcharrison, line 47: This is not ideal, I think, because it doesn't check that the `ImportError` corresponds specifically to `music21`; it could be another package that is missing;

benchmarks/notesize.py
```diff
@@ -0,0 +1,111 @@
...
+from amads.core import Note
+
+"""
+How big is a note?
+"""
+import sys
+
+
+def get_total_size(obj, seen=None):
+    """Recursively find the total memory size of an object, including its attributes."""
```
pmcharrison, line 26: I'm sure people have created specialised packages for quantifying object sizes, isn't it going to be errorprone to write your own? e.g. https://pypi.org/project/objsize/

demos/durdist2.py
```diff
@@ -9,7 +9,7 @@
 import matplotlib.pyplot as plt
 
 from amads.algorithms import duration_distribution_2
-from amads.io import partitura_midi_import
+from amads.io import music21_midi_import
```
pmcharrison, line 12: Might it be better to use a more generic `amads_midi_import` (or my preferred `import_midi`) in these demos instead?

tests/utils_test.py
```diff
@@ -10,7 +10,7 @@ def setUp(self):
         """Set up the test case with example music files"""
         # Example music files
```
pmcharrison, line 11: To get the CI working again I think you should add "music21" to pyproject.toml in the `test` dependencies section.

rbdannenberg: music21 dependency will be there in my next push

amads/io/scoreread.py
```diff
@@ -0,0 +1,119 @@
...
+        return xml_import(filename, show)
+    else:
+        raise Exception(
+            "Could not find a MusicXML import function. "
+            + "Preferred subsystem is"
+            + str(preferred_reader)
+        )
+
+
+def amads_midi_import(filename, show=False):
```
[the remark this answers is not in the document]
rbdannenberg: OK. Original intent was, since we had specific `pt_midi_import` and `m21_midi_import`, to offer a clearly (?) generic `amads_midi_import`. I'm changing to `import_midi` (see also notes on noun_verb vs verb_noun naming scheme in another comment.)

amads/io/scoreread.py
```diff
@@ -0,0 +1,119 @@
...
+        print(f"{preferred_reader} is available, using it instead.")
+        return True
+    else:
+        print(f"{preferred_reader} is not available either.")
+        preferred_reader = original_reader
+        return False
+
+
+def _check_for_subsystem(_try_alternate: bool = True) -> bool:
+    """Check if the preferred reader is available. If so, install it.
```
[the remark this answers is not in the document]
rbdannenberg: Right - probably I didn't sufficiently edit a copilot suggestion.

demos/durdist2.py
pmcharrison: Interesting thoughts! In my experience the `noun_verb` approach often ends up working fine even with OOP, because typically the object (i.e. the instance of the class) is the thing calling the method. e.g.

```py
reader = MIDIReader()
reader.read_midi("file.mid")
```

amads/io/scoreread.py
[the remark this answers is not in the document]
rbdannenberg: It is fixed now.

rbdannenberg:
It's been a long slog examining MIDI files and input. Final conclusion is both Music21 and Partitura are broken. I wrote a 3rd Importer using PrettyMIDI which does much less than others but seems to do it correctly. Also runs all 190 tests in about 7 seconds! @pmcharrison: I think the sliding window code had some floating point issues that would never show up with power-of-two sizes due to exact arithmetic, but in the general case could be a problem. I did a lot of work taking a more defensive-programming approach, adding a lot of explanation. It passes tests, but see what you think. This pull request is not the best place to change that code, but the old code was failing tests (in hindsight, maybe for a completely different reason), so this is the branch where I changed it.

pmcharrison:
Thank you Roger -- I appreciate the effort taken to update that code. I will look properly when I get a moment, but might be a week or so. If you prefer not to get bogged down waiting for me then you're welcome to merge and I can have a look later.

pmcharrison:
Hi @rbdannenberg, looks like you merged it but the tests were still failing?

rbdannenberg:
@pmcharrison Yes, I thought some unrelated tests were failing before and after the merge. All tests pass on my personal machine.

pmcharrison:
Sorry, you're absolutely right, the test failures are unrelated! I'll have a further look soon.

PR #121, opened 2025-04-29
Fix issues creating docs

### LINKED ISSUE #119: `sphinx-build` silently failing when importing modules
In the latest documentation build (https://github.com/music-computing/amads/actions/runs/14733349956/job/41353050746) we're silently failing to import two modules --- `amads.harmony.root_finding.parncut_1988` and `amads.time.npvi`. In the latter case, this file should have been removed by #117, but somehow still seems to be getting found by `sphinx`.

This means that the corresponding documentation for these modules are not showing correctly in the documentation build. Any ideas @pmcharrison @MarkGotham ?

HuwCheston:
This is currently a stop-gap solution that fixes #119 for `time.npvi` and `harmony.root_finding.parncutt_1988` not being included in the documentation by changing these to `time.variation` and `harmony.root_finding.parncutt` inside `index.rst`.

As I mention [in this comment](https://github.com/music-computing/amads/issues/119#issuecomment-2839407227), I think we should find a way to move away from explicitly defining every package inside `index.rst`.  It doesn't seem that refactoring tools like the one in PyCharm catch these changes properly when the filename of a `.py` file is changed, which makes debugging these issues quite challenging.

I think the solution would be to replace some of the explicit definitions inside `index.rst` with something like the following formulation:

```
General algorithms
------------------

.. autosummary::
   :toctree: _autosummary
   :caption: General algorithms:
   :recursive:

   amads.algorithms
```

### DIFF
--- a/amads/time/variability.py
+++ b/amads/time/variability.py
@@ -58,66 +58,65 @@ def normalized_pairwise_variability_index(
     durations: Iterable[float],
 ) -> float:
     r"""
-     Calculates the normalised pairwise variability index (nPVI).
+    Calculates the normalised pairwise variability index (nPVI).
 
-     The nPVI is a measure of variability between successive elements in a sequence.
+    The nPVI is a measure of variability between successive elements in a sequence.
 
-     The equation is:
+    The equation is:
 
-     .. math::
+    .. math::
 
         \text{nPVI} = \frac{100}{m-1} \times \sum\limits_{k=1}^{m-1}
         \left| \frac{d_k - d_{k+1}}{\frac{d_k + d_{k+1}}{2}} \right|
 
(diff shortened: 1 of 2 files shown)


pmcharrison:
Thanks @HuwCheston! Long term, we should have two separate parts to the documentation: (a) a manually curated overview, including discursive introductions and tutorials, and (b) automatically generated API documentation. The current implementation that I setup is a stopgap that doesn't properly achieve either goal. Hopefully we'll get there in the coming months though.

pmcharrison (approved):

HuwCheston:
Ok, in that case @pmcharrison id suggest we merge these changes so the docs can build properly, then keep @MarkGotham 's issue open to keep track of the long term plans?

pmcharrison:
Yep

PR #126, opened 2025-06-06
Implemented midi2hz

isaacLi2005:
This implements the midi2hz function. It converts midi note numbers to a hz measurement. It is very structurally similar to the already-impelemted hz2midi function.

### DIFF
--- a/amads/pitch/midi2hz.py
+++ b/amads/pitch/midi2hz.py
@@ -1,0 +1,38 @@
+"""
+Provides the `midi2hz` function.
+
+Original doc: https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=6e06906ca1ba0bf0ac8f2cb1a929f3be95eeadfa#page=50
+"""
+
+def midi2hz(midi):
+    """
+    Converts a midi note number to the corresponding frequency in hz. 
+    Validates input to make sure that the input is non-negative. 
+
+    Parameters:
+    midi (float or a list of floats): The midi note number or a list of midi note numbers. 
+
+    Returns:
+    hz (float or a list of floats): The corresponding hz measurements for the midi note numbers.
+
+    Raises:
+    ValueError: If any midi note number is negative. 
+    """
+    
+    def convertMidi2Hz(singleMidiNoteNumber):
+        """
+        A helper function that turns one midi note number, input as a single float, and returns
+        the hz measurement as a float; also validates that the input midi is nonnegative. 
+        """
+        if singleMidiNoteNumber < 0:
+            raise ValueError(f"The midi note number must be nonnegative, got {singleMidiNoteNumber}")
+
+        return 440 * (2 ** ((singleMidiNoteNumber - 69) / 12))
+    
+    if isinstance(midi, list):
+        result = []
+        for midiNote in midi:
+            result.append(convertMidi2Hz(midiNote))
+        return result
+    else:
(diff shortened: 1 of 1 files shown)


rbdannenberg (changes requested):
Just for the record: we reviewed in our meeting today. Awaiting some changes to documentation and addition of examples to auto-generate simple tests.

amads/pitch/midi2hz.py
```diff
@@ -0,0 +1,38 @@
...
+    midi (float or a list of floats): The midi note number or a list of midi note numbers. 
+
+    Returns:
+    hz (float or a list of floats): The corresponding hz measurements for the midi note numbers.
+
+    Raises:
+    ValueError: If any midi note number is negative. 
+    """
+    
+    def convertMidi2Hz(singleMidiNoteNumber):
```
pmcharrison, line 22: Defining a function inside a function incurs unnecessary performance overhead, I wouldn't do that

amads/pitch/midi2hz.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        A helper function that turns one midi note number, input as a single float, and returns
+        the hz measurement as a float; also validates that the input midi is nonnegative. 
+        """
+        if singleMidiNoteNumber < 0:
+            raise ValueError(f"The midi note number must be nonnegative, got {singleMidiNoteNumber}")
+
+        return 440 * (2 ** ((singleMidiNoteNumber - 69) / 12))
+    
+    if isinstance(midi, list):
```
pmcharrison, line 32: Personally I don't think we should put vectorisation code like this in our functions, we should leave it to the user.

amads/pitch/midi2hz.py
```diff
@@ -0,0 +1,38 @@
...
+    Raises:
+    ValueError: If any midi note number is negative. 
+    """
+    
+    def convertMidi2Hz(singleMidiNoteNumber):
+        """
+        A helper function that turns one midi note number, input as a single float, and returns
+        the hz measurement as a float; also validates that the input midi is nonnegative. 
+        """
+        if singleMidiNoteNumber < 0:
```
pmcharrison, line 27: Are you sure this check is worthwhile? I don't see anything conceptually invalid about a negative MIDI number.

amads/pitch/midi2hz.py
```diff
@@ -0,0 +1,38 @@
...
+
+        return 440 * (2 ** ((singleMidiNoteNumber - 69) / 12))
+    
+    if isinstance(midi, list):
+        result = []
+        for midiNote in midi:
+            result.append(convertMidi2Hz(midiNote))
+        return result
+    else:
+        return convertMidi2Hz(midi)
```
pmcharrison, line 38: Hmm, I just noticed that the code structure is basically the same as the pre-existing `hz2midi` function, so I guess it is reasonable to be consistent with that.

rbdannenberg: But we can improve both - I suggested this as a "starter" project just to go through the process, and it would be good to end up with some perfected models for documentation, style, interface, citations, etc.

rbdannenberg (changes requested):
I agree with Peter on nested functions and trying to keep style consistent - can you just move convertMidi2Hz out of midi2hz() and rename to _convert_midi_to_hz()? The underscore will "hide" the function even though it is at the module level.

PR #127, opened 2025-06-18
Adding combcontour function

ryfz081:
Adding the miditoolbox functions combcontour and concur. 

Combcontour generates an nxn matrix from a monophonic score of n notes where the i,jth index is 1 if note i is higher than note j and 0 otherwise. Concur calculates the proportion of concurrent onsets in a score.

I created tests for combcontour and concur in the tests folder.

Not sure if my concur function is working as intended, however I am also unsure if the MATLAB miditoolbox is correct.


amads/time/concur.py
```diff
@@ -0,0 +1,42 @@
+import numpy as np
+from ..core.basics import Note, Score
+from ..algorithms.nnotes import nnotes
+
+def concur(score: Score, threshold: float = None):
+    """
+
```
MarkGotham, line 7: Please can you add an explanation of this function here (above the args) like you did with the other one

ryfz081: After talking to Professor Tuomas about this function he said we should probably just not write this function since it isn't necessary.

ryfz081:
> Hi @ryfz081 is this ready?
> 
> When you say you're not sure if "concur function is working as intended" nor ... "if the MATLAB miditoolbox is correct." do you have some reference values?
> 
> I think we'd accept a PR if any diverges that you can't resolve are explicitly identified.
> 
> With that clarification, it's a work in progress and any users are forewarned. Without it, users may be misled.
> 
> Let me know (with a tag @ ) and I'll be happy to review any changes.

Issue #78, opened 2025-03-06
R as dependency


MarkGotham:
Hi @Dav8Circle and @pmcharrison ,

If I'm reading this right, the `melsim` module is a "Python wrapper for the R package" and appears to require the user to have R running locally and rpy2 installed.

Is that temporary? I.e., is the plan to re-implement the functionality locally?

pmcharrison:
Hi @MarkGotham -

Yes, it's currently an R wrapper and requires R/melsim/rpy2 to be installed. It's an optional dependency, which means it will only be installed if you explicitly request it so as to work with melsim.

Yes, the long-term plan would be to port it to Python. Currently the target package is under active development though so it would be good to let the dust settle there first.

MarkGotham:
From @Dav8Circle on #129 

"
something to do with rpy2 has changed in a way I presently can't account for. I've got a hacky fix that I'm using in my personal repo right now if you want me to push that over, or if you're happy to wait a bit, I'm planning on a more robust fix in July.
(https://github.com/Dav8Circle/PhDMelodySet/blob/d27d70aa9276ffb74a33ed19fa6a5953a3a7f381/Feature_Set/melsim_wrapper/melsim.py)
"

Thanks for this Dave.

Do you intend to continue to use the R dependency and wrapper, or re-implement in python (as per short-term / long-term discussion above)?

Dav8Circle:
Hi Mark, the long term plan remains to reimplement in python, but we still aren't sure of when this work should be achieved. Peter and I spoke to Seb in person a couple of weeks ago, and he said there are some meetings happening between his collaborators to formalise the package a bit more - though we'll have to wait a while to see the results of these conversations, I suppose. 

I'll be talking to Klaus in a week or so for my interview, if time allows I'd like to ask him what his thoughts are, too, but still TBD. 

tl;dr we'll be sticking to the wrapper for now - I'll put in a PR for the quick fix, and I'll get a proper reimplementation done at some point once I have a better picture of where melsim is in its development cycle. Hope that sounds okay?

PR #131, opened 2025-06-29
Update to Melsim

Dav8Circle:
as discussed in #78 (@MarkGotham, @pmcharrison), here is my temporary fix to the melsim wrapper - not by any means the nicest code I've ever written, but it gets the job done for now. Happy to develop on this as needed, but would obviously prefer to invest time instead into writing a good reimplementation from scratch in future.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -84,4 +84,4 @@ jobs:
 
     - name: Run melsim tests
       run: |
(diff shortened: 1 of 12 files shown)


pmcharrison:
hi @Dav8Circle, this PR looks quite strange, I think maybe you didn't branch off the latest master, so you are missing lots of recent changes?

PR #132, opened 2025-07-01
Temporary melsim fix

dmwhyatt:
A proposed temporary measure to fix melsim until I have time to write a proper Python port. This should perform the same as the old version that used `rpy2`, but with a slightly different interface: 

`get_similarity_from_scores` is for comparing two `Score` objects using a single method and single transformation.

`get_similarity_batch` is for comparing multiple scores over multiple methods and transformations.

See `demos/melsim.py` for example usage.


demos/melsim.py
```diff
@@ -0,0 +1,208 @@
+"""
+Example usage of the melsim module.
+"""
+
+from amads.core.basics import Score
+from amads.melody.similarity.melsim import (
+    check_r_packages_installed,
+    get_similarity_batch,
```
pmcharrison, line 8: Proposal: `get_similarity` and `get_similarities`

PR #134, opened 2025-07-17
Syncopation metric examples and related

MarkGotham:
TODO: replicate reported results (currently different)

### DIFF
--- a/amads/time/meter/syncopation.py
+++ b/amads/time/meter/syncopation.py
@@ -54,6 +54,9 @@ class SyncopationMetric:
         self, onset_beats: Optional[list] = None
     ) -> float:
         """
+
+        TODO: WIP - does not currently replicate answers in the literature; further investigation to follow
+
         The weighted note-to-beat distance measure (WNBD)
         measures the distance between note starts and
         records the traversing of beats, and the distance to the nearest beat.
@@ -78,7 +81,7 @@ class SyncopationMetric:
         --------
         We use the example of the son clave
         (also available from the `meter.profiles` module),
-        adpating to match presentation in the literature.
+        adapting to match presentation in the literature.
 
         >>> son = [1, 0, 0, 1, 0, 0, 1, 0, 0, 0, 1, 0, 1, 0, 0, 0]
         >>> onset_beats = vector_to_onset_beat(vector=son, beat_unit_length=4)


MarkGotham:
Ah. Got it (or at least part of it):
- off-by-one issue in the note count (sum of n distances notes and n+1 notes)
- trial and error to work out what they think the "beat" is in each example (clue: the time signatures vary but not reliably with their notion of beats)

G2g, Agatha Christie is on the phone.

MarkGotham:
For the 12-element cases most are down by 3/7 on the quoted values

MarkGotham:
WNBD-16

MarkGotham:
Tagging @pmcharrison in case you want to weigh in on the draft Vector class.

rbdannenberg:
@MarkGotham I looked at the vector class and it doesn't look like something that filling an obvious need given Python tuples, lists, NumPy arrays, etc. But I assume this development was driven by some applications in analysis implementations, which is fine. The use and terminology of indicator and interval vectors is new to me, and I can see that they might make it easier to explain what an algorithm does and warrant the introduction of these as independent types, so I didn't have anything to say.  However, I think we both agree with Peter's comments about subclassing and inheritance.

MarkGotham:
> @MarkGotham I looked at the vector class and it doesn't look like something that filling an obvious need given Python tuples, lists, NumPy arrays, etc. But I assume this development was driven by some applications in analysis implementations, which is fine. The use and terminology of indicator and interval vectors is new to me, and I can see that they might make it easier to explain what an algorithm does and warrant the introduction of these as independent types, so I didn't have anything to say. However, I think we both agree with Peter's comments about subclassing and inheritance.

Great, thanks. Personally I tend to get a bit confused between all the different things a tuple might mean, so I can see the motivation for making the user (... myself ;) ...) set it out more clearly.

Why don't I make something like this (OO), but also leave the "flat" functions in and we can see which we find more natural to work with (again on a different PR). 

As for terms, "indicator" is well established outside of music, I think. The "interval vector" is one that's unusually well established within music (at least music theory). The challenge there is that folks will associate it with the case of a vector for summarising the full interval content; so the alternative with only adjacent interval needs careful naming. I'll have a look and see if there's any consensus around that that I'm not remembering. Same for "wrapped" which could be called "cyclic" or "rings", or maybe "ouroboros" ;)

I remember talking to Bob Morris about something related in the context of preparing the 12-tone serial row code. If memory serves, there seemed to be (yet another) Anglo-American language divide ;)

PR #136, opened 2025-07-26
Only run codecov if the tests pass

pmcharrison:
Previously, failing tests would mean that the codecov step runs and fails, which confuses people when debugging. I've now changed this behaviour. I've also made an effort to fix doc builds.


MarkGotham:
Fab thanks!

Maybe an issue to remind us to restore the melsim in due course

... or maybe we have enough issues on that topic already ;) ...

pmcharrison:
Yeah, I don't think we need another reminder haha!

PR #140, opened 2025-08-14
refactor: optimise vector operations with numpy in vectors_sets

cucuwritescode:
Hi, thank you for the toolkit :)
the PR addresses TODO comments in amads/core/vectors_sets.py requesting numpy optimisations for better performance:
  - scalar_multiply: now uses numpy array multiplication instead of list
  comprehension
  - weighted_to_indicator: now uses np.where instead of manual iteration
  
both changes maintain the same interface and behaviour and pass existing tests/doctests.

I’m interested in contributing further improvements to the codebase, I noticed several unimplemented methods that could use attention if you'd like help with those.


MarkGotham:
Hi @cucuwritescode great to have you on board!

Yes please to other methods!

Please let us know which you're planning to implement either in writing, or with an initial draft PR so we avoid duplication.

PR #141, opened 2025-08-19
Adding key finding functions

ty-kcs:
## Things we've done
- Changed data attribute values in [profiles.py](https://github.com/music-computing/amads/blob/main/amads/pitch/key/profiles.py) from using tuples of floats to represent pitch-class distributions to using specific PitchProfile objects that inherit from core/[distribution.py](https://github.com/music-computing/amads/blob/main/amads/core/distribution.py)
- Implemented key_cc which supports all the key profiles from [profiles.py](https://github.com/music-computing/amads/blob/main/amads/pitch/key/profiles.py) which contains transpositionally non-equivalent pitch profiles.
- From key_cc, we have provided a functionally equivalent port of [kkcc](https://github.com/miditoolbox/1.1/blob/master/miditoolbox/kkcc.m) from the miditoolbox version.
- From key_cc, we have also ported some of the functions that were dependent on kkcc, namely [kkkey](https://github.com/miditoolbox/1.1/blob/master/miditoolbox/kkkey.m), [transpose2c](https://github.com/miditoolbox/1.1/blob/master/miditoolbox/transpose2c.m), [keymode](https://github.com/miditoolbox/1.1/blob/master/miditoolbox/keymode.m), and [max_key_cc](https://github.com/miditoolbox/1.1/blob/master/miditoolbox/maxkkcc.m). 
  - In order to support all the pitch profiles as well as future pitch profiles, kkkey, keymode, and max_key_cc were changed to be dependent on key_cc instead. As for transpose2c, the original function only transposed to C-major/minor, so we opted for using the kkcc version instead, although this could be changed relatively simply.

## Things that could be improved (in a future pull request)

- Pcdist1, as well as the distribution functions that are not duration based, currently returns a list of 12 floats. But, we should make it return a distribution object from core/[distribution.py](https://github.com/music-computing/amads/blob/main/amads/core/distribution.py). 
  - By using distribution object, it is easier to do visualization with less code clutter
  - It is better practice to use distribution object when we can to make our codebase more consistent.
-  Transpose2c takes a key profile instead of a profile string, and uses key_cc instead of kkcc as a result. However, we restricts attribute names when calling key_cc to ‘major’ and ‘minor’ instead.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,16 +58,18 @@ jobs:
       with:
         python-version: "3.11"
 
-    - name: Setup R dependencies
-      uses: ./.github/actions/setup-r
+#    - name: Setup R dependencies
+#      uses: ./.github/actions/setup-r
 
     - name: Install Python packages
       run: |
-        pip install -e .[test,melsim]
+#        pip install -e .[test,melsim]
+        pip install -e .[test]
+        pip install typing_extensions
 
-    - name: Run melsim tests with coverage
-      run: |
-        python -c "from amads.ci import run_ci_group_tests; run_ci_group_tests('tests_melsim')"
+#    - name: Run melsim tests with coverage
+#      run: |
+#        python -c "from amads.ci import run_ci_group_tests; run_ci_group_tests('tests_melsim')"
 
     - name: Upload coverage to Codecov
       if: success()


amads/core/distribution.py
```diff
@@ -25,6 +25,8 @@
 DEFAULT_BAR_COLOR = "skyblue"
 
 
+# legit think distribution class is too ambitious and contains too much information...
+# hell even
```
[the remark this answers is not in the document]
MarkGotham: Please review this and, while a bit of character is fine, also tone down "legit" and "hell".

MarkGotham:
Thanks!

Looks good overall.

A few things:
- Please review copilot's comments above.
- Don't worry about float vs distribution object. Use distribution object if it actively helps; float is fine.
- I'll update from main now which should see the PR pass tests.

@rbdannenberg is this one you've reviewed / been supervising offline?

amads/core/distribution.py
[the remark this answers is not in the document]
diwangislucky: Sorry that was my fault. Will fix.

rbdannenberg (commented):
I made some minor changes but ran into a couple of major things you should either work on or we should discuss:
1. (easy) Documentation should follow numpy docs conventions. Not all of AMADS is formatted like this, but it should be.
2. (harder) I thought asymmetrical distributions would be 2-D distributions rather than 12 1-D distributions. I thought the most useful API for users would be something like: find_the_key(distribution, profiles=("QuinnWhite.major_assym", "QuinnWhite.minor_assym")) and you would get back a tuple like ("QuinnWhite.major_assym", 5) meaning the Key is F-major.  Having individual names for each key like QuinnWhite.major_assum.C, QuinnWhite.major_assum.C#, etc., seems to defeat the purpose of combining these into a list of 12 distributions. Maybe I missed some detail, but we need to think about how this is all going to be used 99% of the time, and work backwards to a design that makes it easy and makes the code easy to understand.
3. Documentation needs to be VERY careful about transpositions for the asymmetry case. It looks like the profiles do not list pitch class C first but rather the tonic first, e.g. for E-maj, the profile represents frequencies of (E, F, F#, ..., Eb) and not (C, C#, ..., B), yet the profile labels are ("C", "C#", "D",...). Probably we should keep the original configuration since I assume the numbers were copied from literature in that order, but if that's the case, the labels should be fixed so if you plot the profile, you get correct labels. It actually makes sense to plot in this order so you could compare different keys as tonic-to-tonic, 3rd-to-3rd, etc., which are comparable numbers, vs C to C, which is not really meaningful.

pmcharrison:
I suppose the utility of (2) would be that it could capture different prior probabilities for different keys?

amads/pitch/key/profiles.py
```diff
@@ -156,19 +156,30 @@ def as_tuple(self, root):
             raise ValueError(f"invalid root {root}")
         assert self.type in ["assymetric-profile", "symmetric-profile"]
         if self.type == "symmetric-profile":
+            # symmetrical case
```
rbdannenberg, line 160: Why does PitchProfile have both `type` and `distribution_type` (inherited from Distribution) attributes? If no good reason, use `distribution_type` and rename from `assymmetric-profile` to `assymmetric_key_profile` and from `symmetric-profile` to `key_profile` (underscores to be consistent with Distribution, and names are just my suggestion: "profile" seems very generic -- isn't every distribution some kind of profile?)

amads/pitch/key/profiles.py
```diff
@@ -156,19 +156,30 @@ def as_tuple(self, root):
...
             raise RuntimeError("not implemented yet!")
         else:
+            # assymetrical case
             raise RuntimeError("not implemented yet!")
 
     def as_matrix(self):
+        """
+        follows the following index ordering for the rows and columns of
+        the returned matrix:
```
rbdannenberg, line 169: rows and columns represent different things. Documentation should say what they mean.

diwangislucky:
I'm so sorry... I just committed a rough draft of a design. I was not explicit in all the components I needed to work with. I had a meeting with @ty-kcs tonight and I wanted to share some thoughts.

amads/pitch/key/profiles.py
diwangislucky: Yeah this is definitely a good observation that I missed. I will change it.

As for the naming convention, I am not particularly sure if it is a good idea for assymetric_key_profile when the main class is a key_profile, so I'm thinking of <prefix>_pitch_profile instead as the custom type name.

amads/pitch/key/profiles.py
diwangislucky: As a follow up, I would like to play around with the design a bit more. While I am aware that an initial design is often times at least 80% of the way there, there is always room for surprises. I love just exploring different things (who knows... might just canonically store the data in a hash-table with explicit pitch strings as something off the top of my head) or introspecting my own implicit premises to see if there are methods.

rbdannenberg:
OK, thanks for your continued work on this, and I know it's still in development. Let me know directly (email) when I should do some review or you have any questions. Github sends a notice about every change or commit, so it's sometimes hard to know what to attend to.

diwangislucky:
@rbdannenberg We are resubmitting our code for review. Since the last review, we have done the following:
* Reworked PitchProfile to support handling asymmetric distributions within the class itself.
* Added rich visualization to be able to visualize Distributions and its inherited subclasses.
* Added class methods in Distribution class to be able to compare and contrast the data in different distributions.
  * Added class method in Distribution to be able to stack multiple Distribution plots vertically.
  * Added class method in Distribution to be able to plot grouped histograms of 1-D distributions of the same type.

amads/pitch/transpose2c.py
```diff
@@ -0,0 +1,48 @@
...
+    string option for kkcc
+
+    Returns
+    -------
+    Score:
+        a copy of the input score transposed to C-major/minor
+    """
+    # a point of optimization is to check for empty case of the score...
+    # this is very hacky...
+    if next(score.find_all(Note), None) is None:
```
rbdannenberg, line 36: I think this is optimizing a case with near zero likelihood while penalizing about 100% of real cases. Is it a problem to pass no notes to kkcc? (kkcc should really protect itself -- maybe just return Nan or zero in those cases)

amads/pitch/transpose2c.py
```diff
@@ -0,0 +1,48 @@
...
+    corr_vals = kkcc(score, profile_name)
+
+    key_idx = corr_vals.index(max(corr_vals)) % 12
+    score_copy = score.deepcopy()
+    for note in score_copy.find_all(Note):
+        keynum, alt = note.pitch.as_tuple()
+        # since Pitches with same alt and keynum are equivalent
+        # and Pitches themselves are considered "immutable" in
+        # our representation scheme
+        note.pitch = Pitch(keynum - key_idx, alt)
```
rbdannenberg, line 47: Is there a reason to try to preserve the original alt? If the original alt is a double-sharp or double-flat, it could end up in the transposed pitch. Alternatively, if keynum - key_idx - alt is not in ABCDEFG, it will be ignored anyway. I haven't come up with any case where the extra work here is necessary, although I could easily be missing something.

rbdannenberg (changes requested):
Basically this looks great. I made a few suggestions.

amads/pitch/key_cc.py
```diff
@@ -0,0 +1,138 @@
...
+import numpy as np
+
+from ..core.basics import Score
+from .key import profiles as prof
+from .pcdist1 import pcdist1
+
+
+def key_cc(
+    score: Score,
+    profile: prof._KeyProfile = prof.krumhansl_kessler,
```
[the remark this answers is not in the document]
diwangislucky: This was the name that was given since this module's inception. We should probably change that.

amads/pitch/keymode.py
```diff
@@ -0,0 +1,57 @@
...
+    key_cc
+
+    Returns
+    -------
+    List[str]
+        filtered list of attribute names
+    """
+    corrcoef_pairs = key_cc(score, profile, attribute_names, salience_flag)
+
+    # I'm too lazy to write my own for loop so please forgive this
```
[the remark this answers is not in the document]
diwangislucky: Not really a problem. I was mostly joking around here -_-. Sorry.

amads/pitch/keymode.py
```diff
@@ -0,0 +1,57 @@
...
+
+    The indices correspond to the following keys in ascending order:
+    0 -> C, 1 -> C#, ..., 11 -> B
+
+    Parameters
+    ----------
+    score (Score): The musical score to analyze.
+    profile (Profile): Relevant key profile to obtain data from
+    attribute_names: List of strings to relevant attribute names
+    within the profile
```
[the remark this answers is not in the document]
diwangislucky: I see. Will clarify the attribute_names list and what exactly it denotes. Will also include some examples here.

amads/pitch/transpose2c.py
diwangislucky: Honestly, I think pitch-shift should have been an in-built function in Score class itself. The way it is presented here seems to be pretty hacky.

That being said, isn't keynum the sole indicator of what the actual pitch (in midi) of the note is, and alt is only used for symbolic notation? So even if I include the original alt, wouldn't the pitch of the note itself still be valid? As for keynum - key_idx - alt, isn't keynum-alt also at risk of being outside of ABCDEFG? However, to my knowledge, has any piece of music realistically use these edge cases frequently to the point where we need to process them differently from ignoring them?

rbdannenberg: Good point: transposition should be a method in Score. I'll add it. That still leaves the problem of alt. When you construct a Pitch, the constructor enforces ((key_num - alt) % 12) in [0, 2, 4, 5, 7, 9, 11]. If not, alt is adjusted on the principle that key_num is the "truth" and alt addresses the problem that different enharmonics can represent the same pitch (at least the same on a piano or in equal temperament). I don't think there are a "natural" or "correct" rules for deciding the enharmonic. For transposition, I think I will do something simple for transposition, e.g. keep the alt if possible, otherwise, use the default. I will also create a transpose-to-C where you specify the original key signature as an integer, e.g. A-major (3 sharps) is +3.  Then, I will strive to change the alts on F, C, G by subtracting one before constructing the new Pitches. Also, I can alter the KeySignature events.

rbdannenberg:
Didn't mean to wait so long. We believe this passes all related tests, although others are a problem. Roger needs to integrate this with some work on mkdocs and core, so it will get more attention.

PR #142, opened 2025-08-19
Update paper.md

MarkGotham:
Here's a proposed statement that works towards clarifying our philosophy. FAO all interested, especially @rbdannenberg and @pmcharrison .


pmcharrison:
Thanks @MarkGotham! Good to specify the idea of a meta-package. Might be wise to be a bit clearer though on what you mean by 'incorporating the code as is'. I think what we mean is using those packages as dependencies, but providing a unified Python API for them, as opposed to (for example) copying and pasting their source code into our package (which is what I might have thought you meant by those original words).

MarkGotham:
Thanks for catching that!

Duly clarified.

PR #157, opened 2026-04-28
Matlab and python comparison tests

diwangislucky:
Minor changes, accompanying a test file for comparing snapshotted matlab results with python results.


amads/pitch/pcdist1.py
```diff
@@ -77,15 +77,18 @@ def pitch_class_distribution_1(
     score = cast(Score, score.merge_tied_notes())
     if weighted:
         score.convert_to_seconds()  # need seconds for duraccent calculation
-    initial_value = 1e-12 if miditoolbox_compatible else 0.0
+
+    initial_value = 0.0
```
rbdannenberg, line 81: Why was this line changed? Previously, there was support for miditoolbox_compatible.

rbdannenberg (commented):
github is complaining about checks failing, which seems to be simply reformatting by black (the auto-format program). I think all this is supposed to be automated; at least when I do "git push", often black and other "pre-commit" stuff runs and often modifies files and aborts the "git push". I think you can run this locally with "pre-commit run --all-files" if it's not automatic.

amads/pitch/pcdist1.py
diwangislucky: This was not correct even in miditoolbox compatibility mode. The summation of pitches in the numerator (not normalized pcd) started at 0.0 and should not have started at 1e-12, while the denominator was normalized by sum(pcd + 1e-12)

diwangislucky: To clarify, with this change, the test result conforms with the matlab snapshotted result. Otherwise, it is off by a bit.

diwangislucky:
> github is complaining about checks failing, which seems to be simply reformatting by black (the auto-format program). I think all this is supposed to be automated; at least when I do "git push", often black and other "pre-commit" stuff runs and often modifies files and aborts the "git push". I think you can run this locally with "pre-commit run --all-files" if it's not automatic.

precommit checks fixed

diwangislucky:
After I requested a review, I realized there is a bunch of stuff here that's outdated. Instead, I guess I wanted to bring to light the two bugs that were fixed in this pull request, mainly the formula for duraccent and the pcdist1 bug in miditoolbox_compatible mode (at least according to matlab cross-testing).

rbdannenberg:
I'm trying to catch up on pull requests. After a few minutes of review, I see *many* lines of new code, but I haven't figured out whether this in new functionality or reimplementation or what. Then I read that this is largely outdated and I should, I guess, understand all the outdated code in order comprehend how and where 2 bugs are fixed, then manually apply those to AMADS. This sounds like a project description rather than a pull request. Would it make more sense for you to create a new pull request based on fixing 2 bugs in the existing codebase?

Issue #153, opened 2026-01-08
`test_crafted_nonempty_melody`


MarkGotham:
Hi @ty-kcs !
cc @rbdannenberg 

I see that:
- your `test_crafted_nonempty_melody` is failing on 3.10 and not on the other versions.
- you mention in-code that you "probably need some help with these tests"

The choice of test task seems strange to me: a whole-tone scale over a full chromatic. Both are highly tonally ambiguous and in some normalisation routines, the result is very similar to the whole-tone alone.

If that ambiguity is somehow the point of the test, can you explain? If not, then please can you test something with a more clearly expected value?

Thanks!

PR #159, opened 2026-05-01
onsetwindow + key tests clear

ty-kcs:
## Summary

Two parts: (1) MIDI Toolbox `onsetwindow` port, (2) clearer key-correlation test melodies ([#153](https://github.com/music-computing/amads/issues/153)).

### Part 1: `onset_window` from MIDI Toolbox

Port of [`onsetwindow.m`](https://github.com/miditoolbox/1.1/blob/master/miditoolbox/onsetwindow.m) 

**Window bounds**
- Default (AMADS): `min_time <= onset < max_time` (half-open)
- Optional `miditoolbox_compatible=True`: `min_time <= onset <= max_time` (closed, matches Matlab)

- `passage`: `Score` or `Iterable[Note]` (same pattern as `window.py` / `salami.py`)
- No `timetype` parameter (per review): callers convert units first with `Score.convert_to_seconds()` / `Score.convert_to_quarters()` when needed. Bounds must match the onset units of `passage`.

**Revised points after initial PR version**
- Removed `timetype`; documented that unit conversion is the caller’s responsibility.
- Added `miditoolbox_compatible` (default `False`) so Matlab’s closed interval can be replicated without changing the AMADS default.

### Part 2: Key-correlation test melodies ([#153](https://github.com/music-computing/amads/issues/153))

- Replaced tonally ambiguous test melodies with a clear C-major diatonic up/down pattern.
- Updated related key tests (`key_cc`, `keymode`, `kkcc`, `kkkey`, `max_key_cc`).
- Confirmed: C major ≈ 0.8580, A minor ≈ 0.7550.


ty-kcs:
sorry, I will make better test for onsetwindow.py

amads/algorithms/slice/onsetwindow.py
```diff
@@ -0,0 +1,90 @@
...
+    ----------
+    passage : Score or Iterable[Note]
+        The musical passage to be filtered.
+    min_time : float
+        Minimum limit of the window (inclusive) in quarters (default) or seconds
+    max_time : float
+        Maximum limit of the window (exclusive) in quarters (default) or seconds
+    timetype (str, optional, default='quarters'):
+        Time unit for calculation:
+        - 'quarters': notes per quarter (default)
```
rbdannenberg: I think you should omit the timetype parameter, see more notes below.

amads/algorithms/slice/onsetwindow.py
```diff
@@ -0,0 +1,90 @@
...
+    >>> len(filtered)
+    2
+    >>> [n.pitch.key_num for n in filtered]
+    [62, 64]
+    """
+    if timetype not in ("quarters", "seconds"):
+        raise ValueError(
+            f"Invalid timetype: '{timetype}'. " "Use 'quarters' or 'seconds'."
+        )
+
```
rbdannenberg: I think it's much faster to convert the entire score to timetype (units) than to convert each note separately. Also, given that you can pass in an iterable and in that case you ignore timetype, I believe the USER should specify units simply by calling Score.convert_to_seconds() or Score.convert_to_quarters() before calling onset_window. It reduces the code and avoids confusion if using iterables.

amads/algorithms/slice/onsetwindow.py
```diff
@@ -0,0 +1,90 @@
...
+                    # Convert from quarters to seconds
+                    onset_time = score.time_map.quarter_to_time(onset_time)
+            else:  # timetype == "quarters"
+                if score.units_are_seconds:
+                    # Convert from seconds to quarters
+                    onset_time = score.time_map.time_to_quarter(onset_time)
+
+        # Check if onset is within the window (half-open interval)
+        if min_time <= onset_time < max_time:
+            filtered_notes.append(note)
```
rbdannenberg, line 84: I'm worried this is different from Matlab. In other cases, there's an optional parameter `miditoolbox_compatible` that changes behavior but defaults to False. You should do that here -- it's a small difference, but I don't like the idea of changing existing documented behaviors when one of the points of AMADS is to enable experiments and results to be replicated.

rbdannenberg (commented):
Hi Tai, now that I look at this, I realize I looked at it before. Probably you asked me to, so I apologize for telling you to email me when you probably did already and were just waiting for me. I forgot that after posting my comments I was supposed to "Submit review" so you can see it. -Roger

rbdannenberg (approved):
Looks good to me. Thanks for the summary (and implementation).

PR #166, opened 2026-05-23
Add trim function from Matlab MIDI Toolbox

ArnavS08:
This PR ports the trim() function from the Matlab MIDI Toolbox, returning a flattened score without modifying the original input. A unit test is also included.

### DIFF
--- a/amads/algorithms/trim.py
+++ b/amads/algorithms/trim.py
@@ -1,0 +1,45 @@
+"""
+Implementation of the trim() function, which removes leading silence, from the Matlab MIDI Toolbox
+
+Original Document: https://github.com/miditoolbox/1.1/blob/master/documentation/MIDItoolbox1.1_manual.pdf, Page 93
+
+"""
+
+from amads.core.basics import Score
+
+
+def trim(score: Score) -> Score:
+    """
+    Removes all leading silence from a given score such that the first note starts at time 0.0.
+
+    trim() returns a new, flattened score to ensure that the original score is not altered.
+
+    Parameters
+    ----------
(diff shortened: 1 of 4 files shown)


amads/algorithms/trim.py
```diff
@@ -0,0 +1,45 @@
...
+        The score that is to be trimmed.
+
+    Returns
+    -------
+    Score
+        A new, trimmed, and flattened score.
+
+    """
+    # 1. Get a flattened version of the score
+    flat_score = score.flatten(collapse=True)
```
rbdannenberg: collapse will merge notes from different parts into a single part. I think miditoolbox does not have a clear notion of parts, although probably it has MIDI channels that can serve as parts. In any case, I don't see any reason to collapse the score.

rbdannenberg (changes requested):
Great work! It looks perfect except for the comment on collapse. Can you fix and retest? Also, did you retest everything with pytest? This PR is pretty isolated and independent, but it's still a good idea to test everything before you think about changing the main branch. Also, you should check documentation. For trim(), you should add trim as an "algorithm" in mkdocs.yml, and create a 1-line docs/reference/algorithms/trim.md similar to docs/reference/algorithms/nnotes.md. Then, in the repo root directory, run mkdocs serve, open http://127.0.0.1:8000/ and proof-read your documentation.

ArnavS08:
Thank you for the feedback, I fixed the collapse parameter and retested. I've also added trim to mkdocs.yml and proofread all the documentation, in addition to running with pytest now.

PR #167, opened 2026-06-01
Ported concur function from MIDI Toolbox

ArnavS08:
Ports the concur() function from the MIDI Toolbox. Concur calculates the number of distinct onset groups as a fraction of total notes, where onsets within a threshold are grouped together. Includes unit tests and documentation.
  
Note: this is not an exact replication of the original implementation due to inconsistencies.

### DIFF
--- a/amads/algorithms/concur.py
+++ b/amads/algorithms/concur.py
@@ -14,7 +14,8 @@ def concur(score: Score, threshold: float = 0.2) -> float:
     the total number of note onsets. Onsets within a certain threshold beats
     are grouped and counted as a single onset.
 
-    This algorithm and its results are not an exact replication of the Matlab MIDI Toolbox concur function.
+    This algorithm and its results are not an exact replication of the
+    Matlab MIDI Toolbox concur function.
 
     Parameters
     ----------
@@ -33,17 +34,16 @@ def concur(score: Score, threshold: float = 0.2) -> float:
     # 1. Get all the notes in the score
     notes = score.get_sorted_notes()
 
-    # 2. Get the onset times of the notes
-    onset_notes = [note.onset for note in notes]
+    # If there are no notes, return 0.0
+    if not notes:
+        return 0.0
 
-    # 3. Group onsets that are within the threshold of each other
-    for i in range(len(onset_notes)):
-        for j in range(i + 1, len(onset_notes)):
-            if abs(onset_notes[i] - onset_notes[j]) <= threshold:
-                onset_notes[j] = onset_notes[i]
+    groups = 1
 
-    # 4. Use a set to get the distinct onsets
-    distinct_onsets = set(onset_notes)
+    # 2. Iterate through the notes and count how many distinct onset groups there are
+    for i in range(1, len(notes)):
+        if notes[i].onset - notes[i - 1].onset > threshold:
+            groups += 1
 
-    # 5. Calculate the ratio of distinct onsets to total onsets
-    return len(distinct_onsets) / len(onset_notes) if onset_notes else 0.0
+    # 3. Calculate the ratio of distinct onsets to total onsets
+    return groups / len(notes)
--- a/amads/algorithms/concur.py
+++ b/amads/algorithms/concur.py
@@ -10,9 +10,14 @@ from amads.core.basics import Score
 
 def concur(score: Score, threshold: float = 0.2) -> float:
     """
-    Calculates the number of distinct note onset times as a fraction of
-    the total number of note onsets. Onsets within a certain threshold beats
-    are grouped and counted as a single onset.
+    Calculates the number of distinct onset "groups," where group
+    boundaries are the inter-onset interval greater than threshold
+    (in beats or seconds, depending on the current units in the score).
+
+    Equivalently, counts the number of distinct onset "groups", where
+    groups are consecutive IOIs less than or equal to threshold. Note
+    that the time span of a group is arbitrary since there can be any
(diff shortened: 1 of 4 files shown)


amads/algorithms/concur.py
```diff
@@ -0,0 +1,49 @@
...
+from amads.core.basics import Score
+
+
+def concur(score: Score, threshold: float = 0.2) -> float:
+    """
+    Calculates the number of distinct note onset times as a fraction of
+    the total number of note onsets. Onsets within a certain threshold beats
+    are grouped and counted as a single onset.
+
+    This algorithm and its results are not an exact replication of the Matlab MIDI Toolbox concur function.
```
rbdannenberg: I'm trying to keep line lengths to 80. You should wrap long lines. The tools don't change comments automatically, but you'll see wrapping is automatic in the formatting tool. (And if you hate the way it wraps argument lists to one parameter per line + 2 lines for open/close parent, I hate it too. Maybe the only style standard for Python I really don't like.)

amads/algorithms/concur.py
```diff
@@ -0,0 +1,49 @@
...
+    Returns
+    -------
+    float
+        The number of distinct onset groups divided by the total number of notes.
+    """
+    # 1. Get all the notes in the score
+    notes = score.get_sorted_notes()
+
+    # 2. Get the onset times of the notes
+    onset_notes = [note.onset for note in notes]
```
rbdannenberg: This (at least to me) is a non-intuitive way to group notes. This is a great example of the need for computational thinking and complexity analysis. This algorithm's run time is proportional to n^2 where n is the number of notes. But it's computable in time proportional to n.

More importantly, I think it is wrong -- it admits a new group with the time span of the group would exceed 0.2s, but the specification calls for a new group when the IOI exceeds 0.2s. A group (such as a rolled chord) could span any length of time as long as notes occur within 0.2s of each other.

Finally, I'm trying to encourage the use of the data structure we have. Instead of copying every note.onset into onset_notes and writing onset_notes[i], just write notes[i].onset -- I think it's clearer. Yes, in this case you modifiy onset_notes, so making a copy first makes sense, but I think the algorithm can be faster and simpler and correct(!) without the copy or modification.

tests/test_concur.py
```diff
@@ -0,0 +1,49 @@
+from amads.algorithms.concur import concur
+from amads.core.basics import Note, Part, Score
+
+
```
rbdannenberg, line 4: It would be a good idea to make a test case that fails before fixing the algorithm to make the test case pass.

rbdannenberg (changes requested):
I think this is easy to fix, but let me know if you need more input.

amads/algorithms/concur.py
ArnavS08: That makes sense. So currently, my algorithm is comparing to the cluster start as opposed to the previous note.

ArnavS08:
I made the changes to the algorithm. Please let me know if I misunderstood anything or am missing some optimizations/test cases.

amads/algorithms/concur.py
```diff
@@ -0,0 +1,49 @@
...
+
+"""
+
+from amads.core.basics import Score
+
+
+def concur(score: Score, threshold: float = 0.2) -> float:
+    """
+    Calculates the number of distinct note onset times as a fraction of
+    the total number of note onsets. Onsets within a certain threshold beats
```
rbdannenberg: Description is still wrong if you read it literally. Going back to my email, I realize my first statement of the specification was the same. Here's a better description:

> Calculates the number  of distinct onset "groups," where group
> boundaries are the inter-onset interval greater than `threshold`
> (in beats or seconds, depending on the current units in the score).
>
> Equivalently, counts the number of distinct onset "groups",
> where groups are consecutive IOI's less than or equal to 
> threshold. Note that the time span of a group is arbitrary
> since there can be any number of consecutive IOI's less
> than threshold.

Then for the algorithm, whenever you have special code for edge
cases, e.g. `if not notes:...`, suspect that something is wrong.
Here's another approach:

```
notes = score.get_sorted_notes()
prev_onset = -999  # effectively -infinity
count = 0

for note in notes:
    if note.onset - prev_onset > theshold:
        count += 1
    prev_onset = note.onset
return count / len(notes)
```

tests/test_concur.py
```diff
@@ -0,0 +1,53 @@
...
+    Note(parent=test_part, onset=0.0, duration=1.0, pitch=60)
+    Note(parent=test_part, onset=0.1, duration=1.0, pitch=62)
+    Note(parent=test_part, onset=0.15, duration=1.0, pitch=64)
+
+    # All 3 notes are within threshold of each other
+    assert concur(test_score, threshold=0.2) == 1 / 3
+
+
+def test_concur_chained():
```
rbdannenberg, line 38: This is effectively the same as test_concur_all_concurrent(). I think you should change the test to use a threshold of 0.11 so only 2 notes are grouped and the answer is 2 / 3. I say 0.11 instead of 0.1 because 0.1 would be *exactly* 0.25 - 0.15, but with floating point, you should never assume exact arithmetic, so 0.11 gives a clear distinction between the IOIs of 0.15 and 0.1.

rbdannenberg (changes requested):
It's getting there ...

ArnavS08:
I've updated the algorithm and its docstring, along with the concur_chained test.

rbdannenberg (approved):
Looks good!

PR #169, opened 2026-06-14
Extend current quantize function and create a wrapper function for it

ArnavS08:
Implements the quantize function from the MIDI Toolbox. Also extends the current EventGroup.quantize to have an additional dur_divisions parameter.


amads/core/basics.py
```diff
@@ -308,7 +308,7 @@ def onset(self, onset: float) -> None:
         self._onset = onset
 
 
-    def _quantize(self, divisions: int) -> "Event":
+    def _quantize(self, divisions: int, dur_divisions=None) -> "Event":
         """Modify onset and offset to a multiple of divisions per quarter note.
 
         This method modifies the Event in place. It also handles tied notes.
```
rbdannenberg, line 315: Interesting. Nothing is said about ties, so let's say it. Below this line, add:

> When a note is tied, the onsets of each tied notes are quantized, and durations are adjusted to the inter-onset times. The overall duration is quantized according to `dur_divisions` by adjusting the final note's duration. If the quantized duration results in one or more tied notes having zero or negative durations, those notes are removed from the score and the duration of the remaining tied sequence is adjusted to be the original duration quantized according to `dur_divisions`.
> 
> If a note or tied note, after quantization, has a duration of zero, the note is removed.
> 

(Is this consistent with MIDI Toolbox? Do we need a parameter to enable/disable removal of zero-duration notes?)

When you set offset, you are actually calling a method that changes duration, and you mostly want to quantized tied_duration rather than duration. So I think you should set duration rather then onset since quantizing is pretty low-level, and it's easier IMO to update the "real" attributes directly.

The special cases for dur_divisions == divisions seem unnecessary. There are (now) going to be enough special conditions to check regarding tied notes, so you do not want to make things even messier with separate cases.

ArnavS08: I just committed the changes. I realize now I missed answering your question completely while working on the fixes, sorry about that.

The MIDI Toolbox documentation doesn't seem to address the case of zero-duration notes, so I've kept the current behavior the same. Please let me know if you want me to add a parameter to handle this. 

I also just noticed that the MIDI Toolbox default value for dur_divisions is double onset_divisions, ours defaults to the same currently, should I change this to match the MIDI Toolbox default?

amads/core/basics.py
```diff
@@ -343,11 +355,7 @@ def _quantize(self, divisions: int, dur_divisions=None) -> "Event":
             dur_divisions = divisions
 
         self.onset = round(self.onset * divisions) / divisions
-        if dur_divisions == divisions:
-            quantized_offset = round(self.offset * divisions) / divisions
-        else:
-            quantized_duration = round(self.duration * dur_divisions) / dur_divisions
-            quantized_offset = self.onset + quantized_duration
+        quantized_duration = round(self.duration * dur_divisions) / dur_divisions
```
rbdannenberg: I think this is wrong. Shouldn't you compute quantized_duration as a function of self.tied_duration to get the final duration we're trying to achieve. I think quantizing note-by-note is going to introduce some strange cases. A test case would be a zero-duration note tied to a sixteenth tied to a quarter, and dur_divisions = 1. The final result should change the first note duration to 1 and delete the tied-to notes.

amads/core/basics.py
rbdannenberg: I think we should keep the dur_divisions = divisions default, but add a note that for MIDI Toolbox compatibility, you must explicitly specify dur_divisions = divisions * 2. MIDI does not have ties, so zero-duration notes tied to something are not possible, but I think zero duration without ties is possible in MIDI and MIDI Toolbox might quantize duration to zero, leaving duration = 0, so I think we should remove all zero durations from tied note sequences, but if it gets down to a single note, leave it with zero duration. So every tied sequence will be represented by at least one note in the result, although the duration of that note might be zero. OK, I guess at this point, we need to either provide another function to remove zero-duration notes, which could be something you don't want, or keep zero durations, which could legitimately represent a grace note. So I propose calling the new parameter (put it last) filter=False. If True, it means remove all zero-duration notes (after quantization).

ArnavS08:
I updated everything based on the feedback. Please let me know if there are any issues.

amads/time/quantize.py
```diff
@@ -0,0 +1,71 @@
...
+        Default is False.
+
+    Returns
+    -------
+    Score
+        A new, flattened, and quantized score.
+    """
+
+    # deep copy so tied chains are intact when quantize runs
+    score_copy = copy.deepcopy(score)
```
rbdannenberg: Use score_copy = score.copy()  -- we have our own "deepcopy" that's specific to our data structure and avoids making unnecessary Pitch objects.

rbdannenberg (changes requested):
One small fix, but otherwise looks good. I didn't check all the tests, so again, please look at code coverage and see if everything gets executed. If not, add tests. This is *very* tricky stuff, so I think we have to at least test every line with at least one test.

ArnavS08:
Fixed it; I also got to 100% coverage with pytest.

rbdannenberg (commented):
This is out of sync with Event.quantize, but I made further changes there based on code in this PR. In the end, it was easier to manually merge by picking bits and pieces.

rbdannenberg:
Changes in this PR were manually merged and tested, largely because of changes on basics.py side that would have been corrupted by changes here. quantize is now implemented, documented, tested, mostly derived directly from this PR, so I am now closing the PR and (maybe) this branch as obsolete. Thanks for this work!

PR #170, opened 2026-07-01
Extreme function port from MIDI Toolbox

ArnavS08:
This is the extreme function implementation from the MATLAB MIDI Toolbox.


amads/algorithms/extreme.py
```diff
@@ -0,0 +1,62 @@
+"""
+Implementation of the extreme() function from the Matlab MIDI Toolbox
+
+Original Document: https://github.com/miditoolbox/1.1/blob/master/documentation/MIDItoolbox1.1_manual.pdf, Page 61
+
+"""
```
rbdannenberg, line 7: Add yourself as author if you wish.

amads/algorithms/extreme.py
```diff
@@ -0,0 +1,62 @@
...
+            onset_note[onset] = note
+        else:
+            current = onset_note[onset]
+            if method == "high" and note.pitch.key_num > current.pitch.key_num:
+                onset_note[onset] = note
+            elif method == "low" and note.pitch.key_num < current.pitch.key_num:
+                onset_note[onset] = note
+
+    # 3. Build a new score with only the extreme notes
+    result = Score()
```
rbdannenberg: Since flat_score is already a fresh copy, you are free to modify it and return it. So try something like:
```
part = flat_score.content[0]
part.content.clear()  # remove note content
for note in sorted(onset_note.values(), key=lambda n: n.onset):
    part.insert(note)
```
Since notes are sorted, you could also do:
```
part.content = sorted(onset_note.values(), key=lambda n: n.onset)
for note in part.content:
    note.parent = part
```
I think I like the second version best - it's very simple and direct.
These also have the advantage of preserving any info that was added to original notes but not picked up by the Note() constructor (line 55-60).

rbdannenberg (changes requested):
See notes. Almost done.

ArnavS08:
Fixed it, thanks!

rbdannenberg:
It seems like some checks are not running, but after a restart, I got tests_main to pass on v3.11-v3.14. The function looks good now. Good job!

PR #171, opened 2026-07-01
Port dropshortnotes function from MIDI Toolbox

ArnavS08:
This PR ports dropshortnotes. 

I apologize for taking much longer on this one, I finished most of it sometime back but I tried to ensure I wasn't missing anything this time. 

Please let me know if any changes need to be made.


diwangislucky:
I have a question (that I don't know the answer about). Is it possible to make dropshortnotes output a non-flattened score (implementation perhaps inspired by the algorithm pattern shown in duration_distribution_1 implementation in durdist1.py)?

rbdannenberg:
Yes, I think dropshortnotes should definitely retain a full score structure if that's the input. For example, grace notes normally have zero duration, and it would be very reasonable to want to drop grace notes. And since full scores do not rely on rests to fill out the duration (each event carries its own start time and duration), removing a short note is not really a problem in terms of maintaining structure. One thing to be careful about is "short note" should be defined in terms of tied_duration, not duration, so a very short note tied to a longer note should not be altered (at least that's my opinion, unless there's a good use case for removing very short tied notes -- and note that we're working on quantize with Arnav, and that will effectively delete very short notes before a bar line or after a bar line and tied across the bar line.)

amads/algorithms/dropshortnotes.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Removes notes whose total tied duration is less than or equal to a threshold.
+
+    The score is first flattened, which merges any tied note chains into single
+    notes. Notes whose duration is then less than or equal to ``threshold`` are
+    removed.
+
+    Note: this is not an exact port of the MIDI Toolbox function, which uses
+    strictly less than for the threshold comparison. This implementation uses
+    less than or equal to.
```
rbdannenberg: Why? Did we talk about this? I would have expected strictly less than like midi toolbox, as in "drop everything less than 1/8 note". But I suppose this is a problem for dropping zero-duration notes. I think the best approach is to make zero a special case, so the documentation would be something like:
```
    threshold : float
        Duration threshold in beats. Notes with ``tied_duration < threshold`` or ``tied_duration == 0``
        are removed. Use 0 to drop only zero-duration grace notes.
```

amads/algorithms/dropshortnotes.py
```diff
@@ -0,0 +1,55 @@
+"""
+Implementation of the dropshortnotes() function from the Matlab MIDI Toolbox
+
+Original Document: https://github.com/miditoolbox/1.1/blob/master/documentation/MIDItoolbox1.1_manual.pdf, Page 57
+
+"""
```
rbdannenberg, line 7: Feel free to change to (with your name):
```
<small>**Author**: Mark Gotham</small>
"""
__author__ = "Mark Gotham"
```

amads/algorithms/dropshortnotes.py
```diff
@@ -0,0 +1,55 @@
...
+    -------
+    Score
+        A new, flattened score with short notes removed.
+
+    Notes
+    -----
+    To drop notes strictly shorter than a duration *d*, it is recommended to first quantize the score
+    to an appropriate sub-multiple of *d* so that durations slightly less than
+    *d* round up to *d*. Then call ``dropshortnotes`` with ``threshold = d -
+    1.0e-6``.
```
rbdannenberg: Sorry if we talked about this, but this seems like such a hack, and I don't think it's even correct in all cases. I think the real question should be how to drop notes exactly equal to some duration, but I'm willing to assume nobody really needs/wants that.

amads/algorithms/dropshortnotes.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    # 1. Flatten the score
+    flat_score = score.flatten()
```
rbdannenberg: Did we talk about *not* flattening the score?  I think that would be easy: just remove tied chains of notes from the score instead of just the one note. Instead of flatten() (line 44), you can call .copy() (which makes an almost-deep copy appropriate for this case (Pitches are not deep copied because we assume Pitches are *really* immutable.)

rbdannenberg (changes requested):
Aside from comments, this looks good and almost ready. Did you do a coverage test? It's a good way to check how much tests are actually testing.

amads/algorithms/dropshortnotes.py
ArnavS08: We did initially discuss less than or equal to for handling the zero-duration notes, but having a special case instead is better so I'll change that.

amads/algorithms/dropshortnotes.py
ArnavS08: I don't think we discussed this, but that makes sense. I'll switch over to .copy() and also remove the whole tied chain.

ArnavS08:
I've implemented all the changes, I also ran a coverage test through pytest-cov which gave 100%.

PR #175, opened 2026-08-26
[dev] minor bug-fixes that were caught by cross-testing between midit…

diwangislucky:
There's 2 small bugs that were caught when cross-testing between miditoolbox and amads.

The first one (in duraccent) was also mentioned by Anirudh during our meeting.

For the second one, note that the miditoolbox compatible offset (1e-12) was not present in the numerator in any form, and was only used during normalization (so in the denominator).


rbdannenberg (changes requested):
Is there a problem with this code or a problem with tests? This is failing a pcdist test according to github.

diwangislucky:
Everything should be fixed. Also added matlab cross-testing utilities (updated and simplified version of the original PR).

rbdannenberg (approved):
OK.

PR #176, opened 2026-08-27
Adding onset distribution & onset autocorrelation

anirudh-sub:
Implemented onsetdist and onsetacorr from MIDI toolbox.

Onsetdist creates a distribution of note onsets relative to measure onsets. Onsets are quantized (by flooring to the nearest division, to be consistent with MIDI toolbox) and weighted by note duration

Onset autocorrelation calculates score autocorrelation values using quantized note onsets (this time rounded to the nearest division).

Tests are created for both in the test folder

### DIFF
--- a/amads/time/onsetdist.py
+++ b/amads/time/onsetdist.py
@@ -1,0 +1,81 @@
+from amads.core.basics import Note, Score, Measure
+from amads.core.distribution import Distribution
+from amads.core.histogram import Histogram1D
+
+from fractions import Fraction
+
+__author__ = "Anirudh Subramanian"
+
+def onset_distribution(
+    score: Score,
+    quarters_per_measure: int = 4,
+    divisions: int = 4,
+    name: str = "Distribution of note onsets",
+) -> Distribution:
+    """
+    Returns a distribution of onsets of Notes relative to the onset of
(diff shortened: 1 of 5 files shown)


amads/time/onsetdist.py
```diff
@@ -0,0 +1,81 @@
...
+
+    h = Histogram1D(
+        bin_centers=boundaries[:-1],
+        bin_boundaries=boundaries,
+        ignore_extrema=False
+    )
+
+    for m in score.find_all(Measure):
+        for n in m.find_all(Note):
+            h.add_point(n.onset - m.onset, weight=n.duration)
```
rbdannenberg, line 70: I suspect this is wrong: if there are ties coming into a measure, the tied-to note will appear as a Note within the measure. Rather than enumerating Measures and searching for Notes, enumerate all Notes (which will filter tied-to notes) and then use the measure attribute of the note to find its measure. This is fast since the Measure is generally the direct parent and directly linked from the parent attribute.

rbdannenberg (changes requested):
I think there are a couple of problems that should be fixed. Let me know if you have questions or if you disagree -- I might be missing something.

